PR #297, opened 2015-01-05
Always show payments button

adammeyer:


adammeyer:
from Cheryl

adammeyer:
This way additional expenses can be added to a free event

ryancarlson:
:+1:

PR #301, opened 2015-01-13
Error notify set environment to stage if on stage

adammeyer:


adammeyer:
@ryancarlson Can you take a look at this when you get a chance.  I reordered some of the error reporting to easier to look at.  I also added a check so it will let errbit know if it's a stage or production error.

ryancarlson:
:+1:

PR #302, opened 2015-01-14
Upgrade to AngularJS 1.3.8 and angular-bootstrap 0.12.0

OzzieOrca:


adammeyer:
Is there a reason you didn't use 1.3.8?

OzzieOrca:
The Google APIs page didn't list it (https://developers.google.com/speed/libraries/devguide). But it seems to exist. So we can use that.

adammeyer:
There seems to be an issue with the date block.  Looks like angular is trying to convert it into a timestamp instead of keeping it as a string.  Might have to build some logic for that in the blocks directive.

adammeyer:
Great, no errors now!  I still think we should save it as a string with just the date, not a timestamp.  The reason I think that is best is because we don't have to worry about any timezone issues.  Example: I just saved 2015-01-15 5am (because I'm in eastern time).  If someone in UTC -6 pulls up my registration it will show 2015-01-14.  Plus it will make it easier for the CSV export, won't need to run any conversion there.  What do you think?

OzzieOrca:
I think that's fine but Angular now requires the model to be a date object (https://github.com/angular/angular.js/blob/master/CHANGELOG.md#130-rc4-unicorn-hydrafication-2014-10-01). Should we strip out the timezone of the date object? Maybe there's a better way to handle this...

adammeyer:
I like that idea.  Convert it to a date object, but then convert it back to a string before sending it to the api.  I kind of already did this with the datetime picker. (https://github.com/CruGlobal/conf-registration-web/blob/master/app/scripts/directives/datepicker.js).  Not completely the same, but similar. Maybe we could use this datepicker, just turn off the time part?  Maybe something to explore.

OzzieOrca:
I'll look into it. I might not get to it today though. Where's the best place to intercept values before they go into the api? With a watch in the directive? Validation? Somewhere else that I'm not familiar with yet?

adammeyer:
Try to use the datepicker I have above, or something similar.  The current HTML date dropdown that comes down doesn't work in IE, Safari or Firefox anyway, so it would be nice to use a JS plugin for that.

OzzieOrca:
I'm going to leave this PR as is and I think it is ready to be merged. It doens't give any errors and I'm going to implement the datepicker in https://github.com/CruGlobal/conf-registration-web/pull/305. If you could take a look at the validation stuff (https://github.com/CruGlobal/conf-registration-web/pull/303) that would be great.

OzzieOrca:
Merged into https://github.com/CruGlobal/conf-registration-web/pull/303

PR #303, opened 2015-01-14
Event 167: Implement validation with error messages

OzzieOrca:
I have a few questions/issues with this implementation:
- Do any fields besides email, phone, and number need help text? None of the other fields seem to have any validation besides being required (or not)
- ~~There's no way radio buttons can be invalid before the submit button is clicked, correct? I didn't apply this directive to any of the radio buttons. After clicking the submit button, the app doesn't stay on this page to show invalid things (such as required but unpressed radio buttons).~~
- ~~Both in the front end (user filling out form) and the back end (admin editing registration) if a value is marked as invalid, it doesn't save it correctly. The front end seems to just make the value empty but the back end either makes it blank or randomly pulls some old value that is not what I just entered or the last value. Submitting an invalid value twice seems to do the random thing.  I haven't looked into fixing this bullet point much... How should the app handle saving an invalid value? Should the back end not allow submission but the front end allow the submission (and saving) of anything but not let you finish registration without fixing those issues?~~


OzzieOrca:
There's an `allowInvalid` option (https://docs.angularjs.org/api/ng/directive/ngModelOptions) that allows invalid values to be saved to the model but then https://github.com/CruGlobal/conf-registration-web/blob/master/app/scripts/services/validateRegistrant.js must reimplement the validation checks. How strict do we want to be with validation? I think for all the blocks, besides email (requiring it to be at least a@a), phone (10-15 digits), and number (digits only) the only validation being performed is whether the field is empty or not.

OzzieOrca:
Can you guys think of an easy way to tell whether it is the first time the user has been on a page of the registration? I was thinking everything that they didn't fill in should be marked as invalid when they come back to a page. Maybe this should only happen once they get to the review registration page and click edit...

OzzieOrca:
Should the number block allow decimal numbers? The Underscore.js function `_.isNumber(answer)` currently doesn't

OzzieOrca:
I believe this now solves all of the issues on the front end and correctly displays error messages (and red highlighting) on specific inputs when they are invalid and they have either been touched or the user has already viewed the page and is returning to the page.

I'm not sure what's going on with the backend. The registrations page is still not displaying any values until they are edited. Is that migration thing still being worked on? I think edit modal will now show any errors once you click away from the input but allow the admin to save ANYTHING and performs no validation. I'm not sure if that is desired. We should check that though once it is working again.

app/views/components/answerDisplay.html
```diff
@@ -1,9 +1,14 @@
 <ng-switch on="block.type">
   <span ng-switch-when="nameQuestion"> {{answer.value.firstName}} {{answer.value.lastName}}</span>
   <span ng-switch-when="checkboxQuestion">{{getSelectedCheckboxes(answer.value) | joiner}}</span>
-  <span ng-switch-when="addressQuestion">{{answer.value.address1}} {{answer.value.address2}}<br>{{answer.value.city}}, {{answer.value.state}} {{answer.value.zip}}</span>
+  <span ng-switch-when="addressQuestion">
```
adammeyer: Great add.  :+1:

app/scripts/directives/blocks.js
```diff
@@ -111,7 +111,12 @@ angular.module('confRegistrationWebApp')
   .directive('dateQuestion', function () {
     return {
       templateUrl: 'views/blocks/dateQuestion.html',
-      restrict: 'E'
+      restrict: 'E',
+      controller: function($scope){
+        if(angular.isDefined($scope.answer) && !($scope.answer.value instanceof Date)){
+          $scope.answer.value = new Date($scope.answer.value);
```
adammeyer: Can you explain what this is for?  I'm not sure.

app/scripts/controllers/registration.js
```diff
@@ -69,19 +69,12 @@ angular.module('confRegistrationWebApp')
...
 
-    $scope.validateAndGoToNext = function (isValid) {
-      if (isValid) {
-        if (angular.isDefined($scope.nextPage)) {
-          $location.path('/' + $rootScope.registerMode + '/' + conference.id + '/page/' + $scope.nextPage.id);
-        } else {
-          $location.path('/reviewRegistration/' + conference.id).search('regType', null);
-        }
+    $scope.goToNext = function () {
+      $scope.page.viewed = true;  //flag used in show-errors directive to show errors without inputs being touched upon going back to a page
```
adammeyer: This needs to also take into account the registrant.  If you register a second person, all the required fields are red right away.

app/scripts/directives/blocks.js
adammeyer: Will this be needed with the new datepicker?

OzzieOrca: No. That was to fix the errors with the HTML5 date picker since the angular input date directive requires the model to be a date object.

adammeyer: Okay.  Good, because I noticed it was setting my date to 1969 or something like that.

app/scripts/controllers/registration.js
```diff
@@ -69,19 +69,24 @@ angular.module('confRegistrationWebApp')
...
-    $scope.validateAndGoToNext = function (isValid) {
-      if (isValid) {
-        if (angular.isDefined($scope.nextPage)) {
-          $location.path('/' + $rootScope.registerMode + '/' + conference.id + '/page/' + $scope.nextPage.id);
-        } else {
-          $location.path('/reviewRegistration/' + conference.id).search('regType', null);
-        }
+    //setup visited flags array to store visits by a specific registrant to a specific page
+    if(!$rootScope.visitedPages){
+      $rootScope.visitedPages = [];
```
OzzieOrca: Is it ok to use the `$rootScope` for this? I couldn't find a component to put it in that didn't get reloaded when moving between pages.

adammeyer: Yes, that's fine.

adammeyer:
@OzzieOrca I'm still getting error messages with the HTML date picker not being a valid date.  Can you merge your new datepicker into this branch?  Then I'll try again.

OzzieOrca:
@adammeyer It all seems to be working together well. Can you review?

PR #305, opened 2015-01-19
Event 192: Add Datepicker to dateQuestion

OzzieOrca:
Should probably update the backend datepickers to match if this seems like a good solution. The other datepicker didn't natively support attributes for being required and disabled. It also had issues with updating the model when the value was empty (user doesn't choose date and it isn't required). Also this one is responsive.

### DIFF
--- a/app/views/components/answerDisplay.html
+++ b/app/views/components/answerDisplay.html
@@ -4,7 +4,7 @@
   <span ng-switch-when="addressQuestion">{{answer.value.address1}} {{answer.value.address2}}<br>{{answer.value.city}}, {{answer.value.state}} {{answer.value.zip}}</span>
   <span ng-switch-when="textareaQuestion" class="eventhub-linebreak">{{answer.value}}</span>
   <span ng-switch-when="phoneQuestion">{{answer.value | tel}}</span>
-  <span ng-switch-when="dateQuestion">{{answer.value | date : 'mediumDate'}}</span>
+  <span ng-switch-when="dateQuestion">{{answer.value}}</span>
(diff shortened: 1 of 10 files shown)


adammeyer:
The date picker looks great!  A few things I noticed:
1) I think the date should save in a standard format, something like 2015-01-20, instead of "formatted version".  This will allow the admin to easily sort on the registrations screen.
2) Navigating from month to month before a date is set causes an error:
image

app/views/components/answerDisplay.html
```diff
@@ -4,6 +4,7 @@
   <span ng-switch-when="addressQuestion">{{answer.value.address1}} {{answer.value.address2}}<br>{{answer.value.city}}, {{answer.value.state}} {{answer.value.zip}}</span>
   <span ng-switch-when="textareaQuestion" class="eventhub-linebreak">{{answer.value}}</span>
   <span ng-switch-when="phoneQuestion">{{answer.value | tel}}</span>
+  <span ng-switch-when="dateQuestion">{{answer.value}}</span>
```
adammeyer: This line is not needed if no formatting or manipulation is going to take place.

adammeyer:
Also, there needs to be an easier way to change the year.  If the user wants to enter a date 20 years from now, they shouldn't have to click through all the months.  Maybe the text box should be editable? Just an idea.

OzzieOrca:
Let's not use this branch since the AngularJS port of this jQuery library is buggy. Check out https://github.com/CruGlobal/conf-registration-web/pull/307 instead.

PR #306, opened 2015-01-20
Date picker for registration form

adammeyer:


adammeyer:
@OzzieOrca What do you think of this approach?  I used the datepicker that was already being using on the event details page.  Let me know what you think.

OzzieOrca:
I added disabled and required attributes to the datepicker. I'm not sure if required is implemented properly since the model is somewhat disconnected from the input element.

This datepicker doesn't allow the user to not enter a date. If you delete the date from the text field, the model isn't updated and keeps the old date. Also the picker is set to use the current date by default (which is what you get when you call `moment(scope.localModel)` when the `localModel` is null or undefined) but that is never saved to the model until the user interacts with the picker. In this case, the review registration page marks the date as invalid since it seems to be empty.

You removed the time from the formatting which, according to http://eonasdan.github.io/bootstrap-datetimepicker/Options/#format, should remove the time button but it didn't...

If we can make this one work for all of these cases, I think it would be fine.

adammeyer:
@OzzieOrca I re-looked at this branch again today.  Both issues you brought up don't happen on my machine.  Did you run bower install to get the latest version?

OzzieOrca:
@adammeyer I looked at this again and there's a few things issues with it:
- When the model is null, it doesn't save what is displayed (today's date) to the model without the user interacting with it
- When the date is empty (user deletes the text in the input) it saves 'Invalid Date' to the model and then upon loading that again displays today's date since at some point that value was converted to null

I have a stash where I was trying to fix those issues but ended up working on https://github.com/CruGlobal/conf-registration-web/pull/307 and I think I got it working fully. It has a nicer API. I could come back to this if needed but I was not getting anywhere (struggling with multiple $digests happening :) ) and decided to try something else.

adammeyer:
@OzzieOrca I agree.  I'll close this one for now.

PR #307, opened 2015-01-21
Pickadate.js custom directive

OzzieOrca:
@adammeyer what do you think of this? I was using a premade AngularJS directive earlier that had forked and rewritten this same library but it was pretty buggy and didn't seem maintained. I made my own directive using the newest version of the jQuery [pickadate.js](http://amsul.ca/pickadate.js/) library.

I could look into making the date formatted in a slightly more user friendly way but it might take a little bit of work and I wanted to get your feedback first. If we want a better formatted date, it should probably be used on the review registration page too but I haven't figured out how to get both the user friendly value and the value for the server to that page. Pickadate.js has a few features for providing both of those formatted values but they aren't in an easy spot for Angular to grab. It would take some more directive logic.

### DIFF
--- a/app/css/style.less
+++ b/app/css/style.less
@@ -872,6 +872,16 @@ body {
   }
   .full { width: 100%; display: block; }
 
+  //pickadate.js modifications
+  .picker__select--month, .picker__select--year {
+    padding: .2em;
+  }
+  pick-a-date, pick-a-time {
+    .form-control[readonly]:not([disabled]){
+      background-color: #FFFFFF;
+    }
+  }
+
   .timepicker, .datepicker {
     .add-on { cursor: pointer; }
   }
--- a/app/css/style.less
+++ b/app/css/style.less
@@ -237,11 +237,6 @@ main.front-form {
       font-size: 14px;
       line-height: 1.43;
     }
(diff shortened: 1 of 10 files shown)


adammeyer:
I like this one.  Seems to work well.  I would try to combine pickadate.html and dateQuestion.html, one wouldn't be used without the other.

What are your thoughts on having two datepickers? I'm thinking if we can get the one we are currently using to work, that would be best. One less dependency to keep track of.  What do you think?

OzzieOrca:
@adammeyer This seems to be working great now and displays and saves the date in 2 different formats. Let me know what you think. I know it might still be nice to use 1 datepicker so we could continue to try and make the original one work.

adammeyer:
@OzzieOrca This is looking good.  Can you make the calendar icon on the left clickable? (toggle the datepicker)

OzzieOrca:
@adammeyer The icon is a clickable button now. How does this look?

adammeyer:
Works great, but the icon looks disabled.
image
Can that be fixed?

OzzieOrca:
Merged into https://github.com/CruGlobal/conf-registration-web/pull/303

PR #308, opened 2015-01-21
Only update answers when needed

adammeyer:
@OzzieOrca Can you review this?  This will remove the unneeded PUTs to the answers endpoint when a registration page is loaded.


OzzieOrca:
Ya this looks good and can be merged.

This is related to correctly putting answers but unrelated to any changes you made. As I was debugging, I noticed that Angular sets `answer.value` to undefined when an input is invalid. The app PUTs the change from `oldAnswer.value=<some valid value>` to `answer.value=undefined` but then stops PUTting the user input until the field is valid again (because `answer.value` is `undefined` the whole time).

This might be a solution: https://github.com/CruGlobal/conf-registration-web/pull/303#issuecomment-70596691

PR #309, opened 2015-01-22
Event 177

dsgoers:


dsgoers:
@ryancarlson please review

ryancarlson:
Actually @OzzieOrca or @adammeyer could you review this one please?

adammeyer:
@dsgoers I would add a confirm dialog "Are you sure you want to delete this payment".  Just because it's close to the edit button.

@ryancarlson Deleting a payment returns a 400 "Bad Request" and editing a payment amount returns a success, but doesn't save new amount.

ryancarlson:
Hmm, I would check the correct Payment ID is being submitted on the delete.  For the editing, what type of payment was being edited?

adammeyer:
Payment ID is correct.  I can't delete any Credit Card payments.  I was able to delete a Cash payment.  I was also trying to edit a credit card payment.

ryancarlson:
That makes sense.  Credit card payments and refunds cannot be deleted nor edited (other than editing the description).  Since those payments were already run through auth.net, nothing can be changed about them.

Only payments that were entered by an administrator can be edited or deleted.

adammeyer:
Okay.  @dsgoers can you make the appropriate changes to reflect this?  Do not show the amount text box if credit card or delete button.

dsgoers:
OK, that should do it!

adammeyer:
Looks good! 
I would copy the amount down to the second line on credit card payments:
image

Also, add the confirmation dialog on payment deletion.

ryancarlson:
One other thing I would do in this PR..  Refund is only needed now for credit cards.  If any other type is being given back, then the admin can just update the amount.

ryancarlson:
So that "Refund" link that half appears in the screen shot for Cash, Check, Transfer and Scholarship can go away.

adammeyer:
Do you think they would want record of the refund?

ryancarlson:
For transfer and scholarship definitely no.  Those transactions happen after the fact, so editing the amount is fine (per Cheryl).  For cash and check I can only see it useful if cash is given at one point, and then refunded at another point.  Maybe it does make sense to leave it for those two.

app/scripts/controllers/paymentModal.js
```diff
@@ -175,4 +175,12 @@ angular.module('confRegistrationWebApp')
       });
     };
 
+    $scope.deletePayment = function (payment) {
+      if(window.confirm("Do you really want to delete this payment?")) {
```
adammeyer: use single quotes here

app/views/modals/paymentsModal.html
```diff
@@ -99,13 +99,16 @@
...
+                  <button ng-click="deletePayment(payment)"
+                          ng-show="!editPayment && editPayment.paymentType !== 'CREDIT_CARD'" class="fa fa-trash-o"></button>
                 </td>
               </tr>
               <tr ng-show="editPayment.id === payment.id" ng-class="{'active': $even}">
                 <td>{{editPayment.transactionDatetime | date:'MMM d, y h:mm a'}}</td>
-                <td><input type="number" ng-model="editPayment.amount" class="form-control"
-                           ng-show="editPayment.paymentType === 'SCHOLARSHIP' || editPayment.paymentType === 'TRANSFER' "
-                           step="0.01" min="0"></td>
+                <td ng-show="editPayment.paymentType !== 'CREDIT_CARD'" >
```
adammeyer: This is picky, but remove the space after the ng-show tag.

app/views/modals/paymentsModal.html
```diff
@@ -99,13 +99,16 @@
...
+                  <input type="number" ng-model="editPayment.amount"  class="form-control" step="0.01" min="0">
+                </td>
+                <td ng-show="editPayment.paymentType === 'CREDIT_CARD'" >{{editPayment.amount | moneyFormat}}</td>
```
adammeyer: same here

dsgoers:
@adammeyer please review, I think that's everything

adammeyer:
@dsgoers looks good!  Merge when ready.

PR #310, opened 2015-01-22
Send confirmation email checkbox plus other improvements

adammeyer:


adammeyer:
@ryancarlson @duboisej This is for the sendEmailReceipt option.  Should be ready to go.  Please review.

app/scripts/controllers/paymentModal.js
```diff
@@ -66,6 +65,8 @@ angular.module('confRegistrationWebApp')
     $scope.canBeRefunded = function (payment) {
       return payment.paymentType !== 'CREDIT_CARD_REFUND' &&
         payment.paymentType !== 'REFUND' &&
+        payment.paymentType !== 'TRANSFER' &&
```
ryancarlson: scholarship and transfer can't or should not be refunded.

adammeyer: correct.  That's what that line does.

PR #319, opened 2015-01-30
EVENT-180

adammeyer:


adammeyer:
@ryancarlson There isn't much "group" wording currently that I could find.  Let me know if you see other words I should replace.

ryancarlson:
Let's try a couple more little changes:
 - swap the order of "Allow Group Registrations" and "Sub Registration Type"
 - change the title from "Sub Registration Type" to "Family-only Registration Type"
 - change the "Allow Family Registrations" tooltip wording to: "When a user chooses to register as this registrant type, he or she then will be permitted to add any other registrants (e.g. a spouse or child) to the registration.
 - change the "Family-only Registration Type" tooltip wording to: "When checked, this registrant type will not become available or visible to registrants until at least one registrant which has "Allow Family Registrations" checked has been added to the registration (e.g. "Staff" registrant type with "Allow Family Registrations" must be registered before the "Child" registrant type with "Family-only Registration Type" becomes an available option).

PR #320, opened 2015-01-30
EVENT-191

adammeyer:
Add event information to review registration screen


ryancarlson:
Code looks good... if the layout looks good on stage I'm good with this change

PR #323, opened 2015-02-09
Event 173

adammeyer:


adammeyer:
@ryancarlson Can you review?

app/scripts/controllers/editRegistrationModal.js
```diff
@@ -12,45 +13,37 @@ angular.module('confRegistrationWebApp')
...
-        var r = confirm('Are you sure you want to mark this registration as completed?');
-        if (!r) {
+        if (!confirm('Are you sure you want to mark this registration as completed?')) {
           $scope.saving = false;
           return;
         }
         $scope.registration.completed = true;
-        $http.put('registrations/' + registrant.registrationId, $scope.registration).success(function(){
+        $http.put('registrations/' + originalRegistrantObject.registrationId, $scope.registration).success(function(){
           saveAnswer();
```
OzzieOrca: Why does this call `saveAnswer()` and put all the changed blocks after having put the entire registrant on the line before it? I seems line the `$http.put()` already saved the new answers to the server. Same goes for line 42

OzzieOrca:
I think this mostly looks good.

Do we want to hide the information blocks? Or does that matter...?

For one of my events it sometimes takes 17 requests (My event has 17 blocks so I'm assuming a request for each block) and slightly under 3 seconds to move the registrant to a different type and change a single field. That doesn't seem ideal... But other times it doesn't do that. I'll test it more later.

app/scripts/controllers/editRegistrationModal.js
adammeyer: Good catch

adammeyer:
Try again with latest commit and let me know.

PR #324, opened 2015-02-10
Fix pickadate year dropdown selector from being disabled in IE

OzzieOrca:
disabled attribute was also being applied to the year dropdown in IE. @adammeyer can you review?


adammeyer:
:+1: Great! thanks

PR #325, opened 2015-02-10
Fix error on saving editBlockOptionAdvanced modal

OzzieOrca:
I noticed this while going through the modals and cleaning up styling for the responsive branch. If choice.amount is undefined or 0 (not '0'), this function throws an error about not being able to call replace on undefined and doesn't allow the user to save and close the modal. @adammeyer can you review?

### DIFF
--- a/app/scripts/directives/block.js
+++ b/app/scripts/directives/block.js
@@ -106,7 +106,7 @@ angular.module('confRegistrationWebApp')
               };
 
               $scope.save = function (choice) {
-                if(choice.amount == null){
+                if(!choice.amount){
                   choice.amount = 0;
                 }else if(_.isString(choice.amount)){
                   choice.amount = choice.amount.replace(',','');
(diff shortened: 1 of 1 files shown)


app/scripts/directives/block.js
```diff
@@ -106,7 +106,11 @@ angular.module('confRegistrationWebApp')
               };
 
               $scope.save = function (choice) {
-                choice.amount = choice.amount.replace(',','');
+                if(choice.amount == null){
```
adammeyer: Use the isNull function here: https://lodash.com/docs#isNull  Just so it's consistent with other parts of the app.

OzzieOrca:
`_.isNull(undefined)` returns false... I'm trying to catch undefined and null (I've seen it have the value undefined but not the value null)
What about using `choice.amount === undefined` or `_.isUndefined(choice.amount)`? Now that I look at it `Number(null)` returns 0 so that function can handle the null

PR #326, opened 2015-02-10
split first and last name into 2 columns

adammeyer:


adammeyer:
@OzzieOrca can you review?

OzzieOrca:
LGTM. Seems like a better way of sorting :)

PR #328, opened 2015-02-17
Cleanup to EVENT-173

OzzieOrca:
Hey @adammeyer,  I just noticed you already merged EVENT-173. The version you had works fine but I was thinking about some small changes and didn't get to them till today. I made saving individual answers async and cleaned up logic and refactored duplicate code into functions.

It took me a while to figure out what this code was trying to do and I feel like this version is easier to read. Let me know what you think. It's not urgent in any way.

### DIFF
--- a/app/scripts/controllers/editRegistrationModal.js
+++ b/app/scripts/controllers/editRegistrationModal.js
@@ -1,13 +1,12 @@
 'use strict';
 
 angular.module('confRegistrationWebApp')
-  .controller('editRegistrationModalCtrl', function ($scope, $modalInstance, $http, conference, registrantId, registration) {
+  .controller('editRegistrationModalCtrl', function ($scope, $modalInstance, $http, $q, conference, registrantId, registration) {
     $scope.conference = angular.copy(conference);
     $scope.registration = angular.copy(registration);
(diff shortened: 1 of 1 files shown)


adammeyer:
Okay, thanks!  Do you think there is any disadvantages to saving all answers async?  Will that be too many at one time?

adammeyer:
Will the getRegistrantAndClose wait until all answers have been updated?  That would be important.

OzzieOrca:
Yes `$q.all(answersUpdatePromises).then(getRegistrantAndClose);` will wait until all the promises (`$http` requests) have resolved before calling `getRegistrantAndClose`.

The only reason this could be too many at one time would be if the api server can't handle it but browsers are limited as to how many requests can be pending at one time (http://sgdev-blog.blogspot.com/2014/01/maximum-concurrent-connection-to-same.html). Even if the server is overloaded it would just take longer to respond to each request.

http://stackoverflow.com/questions/23931846/wait-for-all-http-requests-to-complete-in-angular-js has a pretty good explaination of waiting until all `$http` requests have completed in AngularJS.

The documentation for `$q.all(promises)` is at the very bottom of this page: https://docs.angularjs.org/api/ng/service/$q

adammeyer:
ok, great.  I think it looks good!  Thanks for cleaning it up.

PR #329, opened 2015-02-17
Showing registration cost only when it's greater than $0.00.

jeedub:
@adammeyer would you take a look at this, please?  I'm just adding a quick angular thing to hide the registration amount if it's zero dollars.  Thanks.


adammeyer:
Put what you added in line 6 instead.  Replace the isConferenceCost() ng-if.

jeedub:
Okay...sounds good.  Thanks.

ryancarlson:
:+1:

PR #330, opened 2015-02-18
Fix issue with pickadate when it is inside a modal

OzzieOrca:
@adammeyer can you review? When the picker is in a modal, it is confined to the modal container and isn't positioned correctly. This issue came up inside the editRegistrationModal.


adammeyer:
:+1:

PR #332, opened 2015-02-19
Use compressed .js and .css files for pickadate

OzzieOrca:
@adammeyer Can you review and merge this? I guess I had used the non-compressed versions for debugging and never switched them


adammeyer:
The non-compressed version for js are preferred.  During the grunt build process it compresses all the js.  The CSS is not, so compressed is great for that.

OzzieOrca:
Ah ok. I had just started wondering about that. I'll keep the compressed css.

PR #333, opened 2015-02-20
EVENT-206

adammeyer:


adammeyer:
@mattdrees @wrandall22 Ready for review. thanks!

wrandall22:
Looks okay to me

app/scripts/controllers/paymentModal.js
```diff
@@ -53,20 +60,39 @@ angular.module('confRegistrationWebApp')
...
+
+      if(transaction.paymentType === 'SCHOLARSHIP') {
+        transaction.scholarship.scholarshipStatus = 'APPROVED';
       }
 
-      if($scope.newTransaction.paymentType === 'SCHOLARSHIP') {
-        $scope.newTransaction.scholarship.scholarshipStatus = 'APPROVED';
+      if(transaction.paymentType === 'CREDIT_CARD'){
+        $http.get('payments/ccp-client-encryption-key').success(function(ccpClientEncryptionKey) {
+          ccp.initialize(ccpClientEncryptionKey);
```
mattdrees: Not a huge deal, but you could initialize ccp earlier, before the transaction processing work happens. This works, but it will probably cause a longer lag in the user experience than is necessary.

app/scripts/controllers/paymentModal.js
```diff
@@ -39,6 +39,13 @@ angular.module('confRegistrationWebApp')
         alert('Transaction amount must be a positive number.');
         return;
       }
+      if ($scope.newTransaction.paymentType === 'CREDIT_CARD') {
+        var validationError = ccp.validateCardNumber($scope.newTransaction.creditCard.number || '');
+        if(validationError){
+          alert('Please enter a valid card number. The ' + validationError + '.');
+          return;
+        }
```
mattdrees: is this check a fall-back, in case the validation in payments.js doesn't work?

adammeyer: No, it's using ccp.js to validate the card number.

mattdrees: That's not quite what I meant. It's clearly using ccp.js here. But you also use it payment.js. Why do you need to do both?

adammeyer: payment.js is for when the user is submitting a payment, the paymentModal.js is for when an admin is manually entering a payment.

mattdrees: ah, I see; thanks

PR #334, opened 2015-02-23
EVENT-166: Force blocks of profile type NAME or EMAIL to be shown for all reg types

OzzieOrca:
@adammeyer can you review? Is the best way to do this to hide the entire tab? We could also just disable the tab so it looks disabled on mouseover and isn't clickable.
https://jira.cru.org/browse/EVENT-166


adammeyer:
Looks good and functions well.  I would go with the disabled tab.  That way the user won't think something is missing.

Don't merge this yet, I'm going to talk with Ryan when he gets back.  I know there are currently events that don't require an email address for some registration types.  I don't know what the official rule will be.  We might just require name.

PR #335, opened 2015-02-25
Split current and pending users

adammeyer:


adammeyer:
@OzzieOrca Can you review when you have a min?

OzzieOrca:
Looks good

PR #338, opened 2015-03-02
discount to registration

adammeyer:


adammeyer:
@ryancarlson Can you review?  This adds the ability for an admin to add a discount.

ryancarlson:
I don't think the API is setup to recognize 'DISCOUNT' as a type, is it, @mattfred ?

adammeyer:
Correct.  It's saving it as Additional Expense.

ryancarlson:
Got it.  The transaction type is what threw me off.  I thought the API was tracking that, but I checked the code and see that it's not.

ryancarlson:
This looks good.  Please test it on stage and verify there before deploying to production.

Also, I would like the API to change the name of the database table and entity and endpoint to be `adjustment` instead of `expense` soon.  The client should update the path it uses too when the API changes. This doesn't have to happen right away, but just so we don't forget about it.

PR #340, opened 2015-03-10
Withdraw/check-in improvements

adammeyer:


adammeyer:
@OzzieOrca can you review?

app/scripts/controllers/eventRegistrations.js
```diff
@@ -386,7 +386,15 @@ angular.module('confRegistrationWebApp')
       }
 
       //update registration
-      $http.put('registrations/' + registrant.registrationId, $scope.getRegistration(registrant.registrationId)).error(function(){
+      var registrationIndex = _.findIndex($scope.registrations, { 'id': registrant.registrationId });
+      var registrantIndex = _.findIndex($scope.registrations[registrationIndex].registrants, { 'id': registrant.id });
+      $scope.registrations[registrationIndex].registrants[registrantIndex] = registrant;
```
OzzieOrca: What does this and the 2 lines above it do? It seems like this was already accomplished in the old version. Does modifying the registrant not update the registrations object? Is it created using pass by value and not reference somewhere? I think JS passes objects by reference but I'm not certain.

OzzieOrca:
This seems to function fine. I guess I'm confused what the "improvements" are. Adding the `$rootScope.loadingMsg` makes sense as an improvement. I'm not sure what the rest does.

You also didn't really give me any explanation of what these changes are for. I just had to read the uncommented code and try to figure out what it does. Short descriptions of why you are changing something would be nice :)

app/scripts/controllers/eventRegistrations.js
adammeyer: I thought it did tool.  But we had some users report that withdraws and checkins not saving.  I think it might be related to this.  When the registration object is updated (with a GET request).  I don't think it is propagating down properly.

PR #341, opened 2015-03-11
Create modalMessage service to display error messages, info messages, and yes/no prompts in modals

OzzieOrca:
Removed all Javascript `alert()` and `confirm` calls. Removed existing modals that fit either the message or yes/no prompt format. @adammeyer can you review? I was planning on doing this sometime but the repeated identical $modal calls in eventRegistrations.js were annoying me.


adammeyer:
Looks good!

PR #342, opened 2015-03-17
Consolidate Validation

OzzieOrca:
Hey @adammeyer I started working on this but had a couple questions. What I have so far is just a different implementation of the validation that is used when the user is viewing the inputs and turns invalid ones red as they fill the form out.

The validation functions are created on the blocks such as https://github.com/CruGlobal/conf-registration-web/compare/consolidate-validation#diff-746ce169ac7ca558d4127ab122be5a20R9. Most of these use the `genericValidators` defined in https://github.com/CruGlobal/conf-registration-web/compare/consolidate-validation#diff-d309ee53729e87610d0c39d229a7fde1R23. They are added to the AngularJS ngModelController validators pipeline in https://github.com/CruGlobal/conf-registration-web/compare/consolidate-validation#diff-e6aa8dfafc5eeb2b5ce1e23d6d242587R12.

The goal is to loop through the blocks as needed and return the value of running all of the validators for that block. This would be used to highlight each invalid block individually on the `reviewRegistration` page and to loop through all the blocks in a conference to determine whether the entire registrant or registration is valid.

Here are my questions:
1. Why is there an `angular.copy()` in https://github.com/CruGlobal/conf-registration-web/blob/master/app/scripts/controllers/registration.js#L27? There is not one in the `reviewRegistration` controller. With this the `$scope.conference` is a different object that the `conference` that is passed into the `validateRegistrant` service and therefore the validators created on the blocks inside `$scope.conference` aren't available in `conference`.
2. Would it be ok to replace the `<span>`s in the switch statement in https://github.com/CruGlobal/conf-registration-web/blob/master/app/views/components/answerDisplay.html with the block directives (such as `<name-question>`) like in https://github.com/CruGlobal/conf-registration-web/blob/master/app/views/components/blockDirective.html? The block directive needs to be run to create the validators. I would then have an ng-if condition in each block directive with one case rendering the inputs and the other case rendering what is currently in https://github.com/CruGlobal/conf-registration-web/blob/master/app/views/components/answerDisplay.html for each block type. We could possibly merge the `showAnswer` and `crsBlock` directives as their functionality is very similar.

How does the implementation look and do you have any thoughts or concerns with using these functions with the rest of the validation functionality?

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -118,6 +118,7 @@
   <script src="scripts/services/unauthorizedInterceptor.js"></script>
   <script src="scripts/services/uuid.js"></script>
   <script src="scripts/services/validateRegistrant.js"></script>
+  <script src="scripts/services/validation.js"></script>
   <!-- endbuild -->
   <!-- build:js scripts/errorNotify.js -->
   <script src="components/hoptoad.js"></script>
--- a/app/scripts/directives/blocks.js
+++ b/app/scripts/directives/blocks.js
@@ -6,7 +6,7 @@ angular.module('confRegistrationWebApp')
       templateUrl: 'views/blocks/nameQuestion.html',
       restrict: 'E',
       controller: function ($scope) {
-        $scope.validators = {
+        $scope.block.validators = {
           firstName : {required: validation.genericValidators.required($scope.block.required)},
           lastName : {required: validation.genericValidators.required($scope.block.required)}
         };
@@ -20,7 +20,7 @@ angular.module('confRegistrationWebApp')
       templateUrl: 'views/blocks/addressQuestion.html',
       restrict: 'E',
       controller: function ($scope) {
-        $scope.validators = {
+        $scope.block.validators = {
           address1 : {required: validation.genericValidators.required($scope.block.required)},
           city : {required: validation.genericValidators.required($scope.block.required)},
           state : {required: validation.genericValidators.required($scope.block.required)},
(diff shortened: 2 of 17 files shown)


adammeyer:
Hey @OzzieOrca - I really like what you've done so far. Looks great!

1) I'm not sure why the angular copy is there, I believe that can be removed.  One things to take into consideration: Modifying the conference object for most users won't be a big deal. But if the user is also an admin for the event and makes changes on the edit details page, the conference object with the validation would be submitted and then rejected by the API.

2) Yes, that sounds good.  When the block directive is run, angular loads all the html views for every block, even if that block isn't used (that is something to be fixed).  I'm assuming that won't happen to display the answer, is that correct?

Thanks!

OzzieOrca:
Cool. Thanks :) I might get to work on it some more this weekend. We've had a bunch of staff meetings this week since it's the end of the quarter.

1. Sounds good. I'll probably have to figure out a way to filter out the `block.validations` object on the eventForm and eventRegistration page (cuz they both use blocks). Or not create them on those pages cuz they probably aren't needed. Although they would probably be useful for the edit registrant functionality.

2. I can't think of a way for that not to happen since the directives in blocks.js need to be run to generate the validators and the ng-switch seems to instantiate all of them and their templates even if they aren't used. Since this is already a problem what do you think of using something like this:
https://github.com/astik/grunt-angular-combine - grunt task to generate a single file with multiple templates
https://github.com/astik/angular-combine - directive to load multiple templates from a single file
Files that match a regex can be grouped into a single file. I haven't tried this yet...

PR #344, opened 2015-03-23
Event 228

ryancarlson:
Try the Facebook JavaScript SDK for facebook logout.

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -196,7 +196,11 @@ angular.module('confRegistrationWebApp', ['ngRoute', 'ngCookies', 'ui.bootstrap'
                   delete $cookies.crsAuthProviderType;
                   delete $cookies.crsPreviousToken;
                   delete $cookies.crsToken;
-                  $window.location.href = headers('X-Facebook-Logout-URL');
+
+                  FB.logout(function(response) {
+                    $location.url('/#');
+                  });
+
                 }).error(function (data, status) {
                   alert('Logout failed: ' + status);
                 });
--- a/app/index.html
+++ b/app/index.html
@@ -131,7 +131,7 @@
         FB.init({
             appId      : '217890171695297',
             xfbml      : true,
-            version    : 'v2.1'
+            version    : 'v2.2'
         });
     };
 
--- a/app/index.html
+++ b/app/index.html
@@ -126,6 +126,24 @@
 </head>
 
 <body ng-class="globalPage.bodyClass" ng-cloak>
+<script>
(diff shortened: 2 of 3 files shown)


ryancarlson:
@adammeyer you can test this out on stage.  can you review please?

app/scripts/app.js
```diff
@@ -192,11 +192,15 @@ angular.module('confRegistrationWebApp', ['ngRoute', 'ngCookies', 'ui.bootstrap'
                  * server side so the users access_token can be fetched to build the log out URL.
                  * after the GET, if successful, then delete the cookies. */
               } else if ($cookies.crsAuthProviderType === 'FACEBOOK') {
-                $http.get('auth/facebook/logout').success(function (data, status, headers) {
+                $http.get('auth/facebook/logout').success(function () {
```
adammeyer: Is this part still needed?

ryancarlson: Yes, it sets that auth token to 'expired' on the server side.

adammeyer: Okay, does something like that need to be done for the relay one too?

ryancarlson: Yeah, it would be a good idea.

ryancarlson: I assume b/c you're asking suggests that it's not currently doing that?

app/scripts/app.js
```diff
@@ -192,11 +192,15 @@ angular.module('confRegistrationWebApp', ['ngRoute', 'ngCookies', 'ui.bootstrap'
...
                   delete $cookies.crsAuthProviderType;
                   delete $cookies.crsPreviousToken;
                   delete $cookies.crsToken;
-                  $window.location.href = headers('X-Facebook-Logout-URL');
+
+                  $facebook.logout().then(function(){
+                    $location.url('/#');
```
adammeyer: If the path is local to the app, you can use $location.path('/'). But both work.

app/scripts/app.js
adammeyer: Correct, the relay logout redirects to https://signin.cru.org.

adammeyer:
Looks good!

app/scripts/app.js
ryancarlson: Okay, I added calls to the API for relay and no auth tokens.  I need to build those methods on the API though (tomorrow) so this won't work on stage yet.. but you could still review the code? :)

adammeyer: Looks good.  Do the endpoints need to different for different auth providers? Could it just be one /auth/expire or something simliar?

PR #448, opened 2016-03-15
GET registration before update of registrant

adammeyer:
@ryancarlson this is a temporary fix until we get the PUT registrant built.  Can you review?

Thanks @dsgoers for the idea!


ryancarlson:
good idea! :+1:

PR #449, opened 2016-03-21
Event 404 - check in permission

dsgoers:
This PR adds check in, a new permission level that is similar to update except users cannot delete registrations or update the event.

### DIFF
--- a/app/scripts/controllers/eventRegistrations.js
+++ b/app/scripts/controllers/eventRegistrations.js
@@ -232,7 +232,7 @@ angular.module('confRegistrationWebApp')
           var registrantIndex = _.findIndex($scope.registrations[index].registrants, { 'id': registrantData.id });
           $scope.registrations[index].registrants[registrantIndex] = registrantData;
         }).error(function(){
-          modalMessage.error('Error: registrant data could be be retrieved.');
+          modalMessage.error('Error: registrant data could not be retrieved.');
           delete expandedRegistrations[r];
         });
       }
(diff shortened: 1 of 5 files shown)


app/scripts/controllers/eventRegistrations.js
```diff
@@ -232,7 +232,7 @@ angular.module('confRegistrationWebApp')
           var registrantIndex = _.findIndex($scope.registrations[index].registrants, { 'id': registrantData.id });
           $scope.registrations[index].registrants[registrantIndex] = registrantData;
         }).error(function(){
-          modalMessage.error('Error: registrant data could be be retrieved.');
+          modalMessage.error('Error: registrant data could not be retrieved.');
```
adammeyer: Thanks for fixing the typos.  Will this endpoint ever return a json error message?  If so, let's add that to this modal.

app/scripts/controllers/eventRegistrations.js
```diff
@@ -275,7 +275,7 @@ angular.module('confRegistrationWebApp')
           $scope.registrants[index] = r;
         });
       }).error(function(){
-        modalMessage.error('Error: registrant data could be be retrieved.');
+        modalMessage.error('Error: registrant data could not be retrieved.');
```
adammeyer: same here

app/scripts/controllers/eventRegistrations.js
dsgoers: I suppose the only times an error would be returned would be if the user has inadequate permissions, but the endpoint requires READ, so that would be rare.  If there was an error in the API, this error is just as good if not better.

I noticed the typo when I was deploying the API locally, so it was temporarily unavailable.  I clicked the edit registrant button and the error message came up.

adammeyer: Okay, then I think it's fine to keep the generic message.

adammeyer:
:+1: Good to merge whenever the API is ready.  Nice job!

PR #450, opened 2016-03-22
Require full permission to edit/save expenses

adammeyer:


dsgoers:
I assume you tested it?  Looks good :+1:

PR #451, opened 2016-03-22
Use registrant endpoint to update registrants

adammeyer:


adammeyer:
@dsgoers This is for the new PUT endpoint

PR #454, opened 2016-03-31
Event 411 - non us staff payments

dsgoers:
Added a third account type for transfer/scholarship payments.  The conferences team wanted a field that can have a longer/shorter number and/or a name and description.  For now, this is OK with them.


adammeyer:
Looks good :+1:

dsgoers:
Thanks.  I'll let you know when the API branch is ready.

dsgoers:
Can you review this again?  Instead of having an extra field for international numbers we use the original account number and don't do validation in the API.

adammeyer:
Looks good :+1:

PR #459, opened 2016-05-26
Explicitly add phantomjs-prebuilt and grunt-cli

OzzieOrca:


package.json
```diff
@@ -18,12 +18,16 @@
...
     "grunt-open": "~0.2.0",
+    "grunt-processhtml": "latest",
     "grunt-rev": "~0.1.0",
     "grunt-usemin": "~2.2.0",
-    "grunt-processhtml": "latest",
-    "karma-phantomjs-launcher": "latest",
     "karma-jasmine": "^0.3.6",
-    "matchdep": "~0.1.2"
+    "karma-phantomjs-launcher": "latest",
+    "matchdep": "~0.1.2",
```
adammeyer: Are we using matchdep?

OzzieOrca: Looks like it. https://github.com/CruGlobal/conf-registration-web/blob/master/Gruntfile.js#L17 I had to look up what it does. I did an `npm install --save-dev` and it reordered some of these dependencies.

adammeyer: okay, thanks!

adammeyer:
👍

PR #460, opened 2016-05-26
Event 352

dsgoers:
Removed readyToProcess and prevent users from attempting partial credit card refunds within 24 hours of the payment (this is also enforced by the API).


app/scripts/controllers/paymentModal.js
```diff
@@ -143,16 +141,45 @@ angular.module('confRegistrationWebApp')
...
         refundedPaymentId: payment.id,
         registrationId: payment.registrationId,
         paymentType: 'REFUND',
-        refundChannel: payment.paymentType,
-        readyToProcess: true
+        refundChannel: payment.paymentType
       };
     };
 
+    $scope.refreshRefund = function (refund) {
```
adammeyer: Can you explain what this function is for?

dsgoers: It determines whether a partial refund is available, and if not, disables the amount window and shows the warning.

app/scripts/controllers/paymentModal.js
```diff
@@ -143,16 +141,45 @@ angular.module('confRegistrationWebApp')
...
+      var paymentToRefund;
+
+      _.forEach($scope.registration.pastPayments, function(value) {
```
adammeyer: You can use the _.find here.

app/scripts/controllers/paymentModal.js
```diff
@@ -311,4 +338,22 @@ angular.module('confRegistrationWebApp')
       var registrationPromoCodes = _.pluck($scope.registration.promotions, 'id');
       return !_.contains(registrationPromoCodes, p.id);
     };
+
+    /*
+    Returns false is payment if payment and refund channel are credit card and current time is less than 24 hours from
+    payment time.  If a refund is issued before the payment settles, authorize.net will do a full refund regardless of amount.
+    */
+    $scope.isPartialRefundAvailable = function(payment, refundChannel) {
```
adammeyer: This doesn't need to be a $scope function if it's not being accessed by the view.  Change this to var isPartialRefundAvailable = function...

adammeyer: Or, you could keep this and remove $scope.onlyFullRefund.  It looks like that is just being set to the same value as $scope.isPartialRefundAvailable.

dsgoers: I added `$scope` to it so I could call it in the tests.

adammeyer: Okay.  It should only be called during the digest cycle.  What do you mean by a lot?

dsgoers:
All changes addressed.

adammeyer:
👍

PR #466, opened 2016-06-20
Update index.html

JasonBuckner:
adjust copyright year to current

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -267,7 +267,7 @@
 
 <footer id="frontfooter" ng-if="globalPage.type === 'registration'">
   <div class="container">
-    <span>&copy; 2014 | Created with Event Registration Tool, powered by <a href="http://www.cru.org" target="_blank">Cru</a></span>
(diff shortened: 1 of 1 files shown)


MeatpieNov:
Resolves https://secure.helpscout.net/conversation/170220690/36752/?folderId=320675

ryancarlson:
Why not create a small service that just grabs the year from the current date?  That way we don't have to update it every year :)

JasonBuckner:
Well, the one I have the most experience with is `<?php echo date("Y") ?>`, but that's not really an option here.

ryancarlson:
@JasonBuckner we like to have someone give an approval to the PR before it gets merged.  this one's trivial, so not a big deal, but in the future you should wait for a 👍  or "Looks good to me" before merging.

JasonBuckner:
Understood. Thanks for the heads up on protocol @ryancarlson .

dougworner:
Why not just copy/paste in JavaScript?

<script language="javascript" type="text/javascript">
     var today = new Date()
     var year = today.getFullYear()
     document.write(year)
</script>

JasonBuckner:
Care to make a Pull request for that @dougworner ?

PR #481, opened 2016-08-02
Angular FB service upgrade

LaelWatkins:
set the Angular FB service to a version that's supported


LaelWatkins:
this isn't necessary.  the scribe lib in the ert api handles calls to facebook.

PR #487, opened 2016-09-12
Event 445

jcsaplar:
These are the first changes that Robin wanted for the view.


jcsaplar:
@hlbraddock This code was just updating the labels to what Robin requested. No logic changed or added.

hlbraddock:
Looks good. +1

PR #491, opened 2016-09-27
Event 381 default options for questions

jayakrishnanv:
@hlbraddock , Added default option functionality. This branch also contains Min / Max Number question functionality as it is linked with the Default functionality. Please test.


hlbraddock:
@OzzieOrca Can you review? As per https://jira.cru.org/browse/EVENT-381

jayakrishnanv:
@hlbraddock @jcsaplar @OzzieOrca Please use #500 instead of this.

PR #494, opened 2016-09-28
Event 380 valid option range for date questions

jayakrishnanv:
@hlbraddock , Implemented Min / MAx range for Date questions, please test.

### DIFF
--- a/app/less/components.less
+++ b/app/less/components.less
@@ -92,3 +92,18 @@ pick-a-date, pick-a-time {
     }
   }
 }
+
+/****** min and max date styles *****/
+.date-alignment{
+    padding-top: 6px;
+}
+.valid-range-msg{
+ min-width:97px;
+ float:left;
+ font-weight:normal;
+ padding-left:15px;
+
+}
+.margin-left-10{
+	margin-left:10px;
+}
--- a/app/less/components.less
+++ b/app/less/components.less
@@ -97,13 +97,3 @@ pick-a-date, pick-a-time {
 .date-alignment{
     padding-top: 6px;
 }
-.valid-range-msg{
- min-width:97px;
- float:left;
- font-weight:normal;
- padding-left:15px;
-
-}
-.margin-left-10{
-	margin-left:10px;
-}
--- a/app/scripts/directives/pickadate.js
+++ b/app/scripts/directives/pickadate.js
@@ -58,8 +58,8 @@ angular.module('confRegistrationWebApp')
         scope.setMinDate();
         scope.setMaxDate();
 
-        scope.$watch('minDate', function (newMinDate, oldMinDate) {
-          if (angular.isUndefined(newMinDate) || newMinDate === '' || angular.equals(newMinDate, oldMinDate)) {
+        scope.$watch('minDate', function (newMinDate) {
+          if (angular.isUndefined(newMinDate) || newMinDate === '') {
              scope.picker.set('min',false);
             return;
           }        
@@ -69,8 +69,8 @@ angular.module('confRegistrationWebApp')
           scope.picker.set('min',dateArray);         
         }, true);
 
-        scope.$watch('maxDate', function (newMaxDate, oldMaxDate) {
-          if (angular.isUndefined(newMaxDate) || newMaxDate === '' || angular.equals(newMaxDate, oldMaxDate)) {
+        scope.$watch('maxDate', function (newMaxDate) {
+          if (angular.isUndefined(newMaxDate) || newMaxDate === '') {
              scope.picker.set('max',false);
             return;
           }
@@ -81,11 +81,9 @@ angular.module('confRegistrationWebApp')
         }, true);
 
         //when date is chosen, update model
-        var onSet = function(){          
-          //scope.$apply(function(){
+        var onSet = function(){
             ngModelController.$setViewValue(scope.picker.get('select', 'yyyy-mm-dd'));
             ngModelController.$setTouched();
-          //});
         };
         scope.picker.on('set', onSet);
 
--- a/app/scripts/directives/pickadate.js
+++ b/app/scripts/directives/pickadate.js
@@ -53,6 +53,26 @@ angular.module('confRegistrationWebApp')
             scope.picker.set('max',dateArray);
           }
         };
+
+         //function to correct current selected date if min and max date changed
+        scope.checkDateRange = function () {
+          if (angular.isUndefined(scope.picker.get('select', 'yyyy-mm-dd')) ||
+            scope.picker.get('select', 'yyyy-mm-dd') === '' || ((angular.isUndefined(scope.maxDate) ||
+              scope.maxDate === '') && (angular.isUndefined(scope.minDate) || scope.minDate === ''))) {
+            return;
+          }
+
+          var date = moment(scope.picker.$node[0].value, 'MMM DD, YYYY');
+          var startDate = moment(scope.minDate, 'YYYY-MM-DD');
+          var endDate = moment(scope.maxDate, 'YYYY-MM-DD');
+          if ((!angular.isUndefined(scope.minDate) && scope.minDate !== '' &&
+            date.isBefore(startDate)) || (!angular.isUndefined(scope.maxDate) &&
+              scope.maxDate !== '' && date.isAfter(endDate))) {
+            scope.picker.clear();
+          } else {
+            scope.picker.$node[0].value = scope.picker.get('select', 'mmm d, yyyy');
+          }
+        };
 		
         //set min and max date
         scope.setMinDate();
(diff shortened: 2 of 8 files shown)


hlbraddock:
@OzzieOrca Can you review? As per https://jira.cru.org/browse/EVENT-380

app/views/components/pickadate.html
```diff
@@ -4,3 +4,8 @@
   </span>
   <input type="text" placeholder="Select a date..." class="form-control" ng-disabled="disabled">
 </div>
+<div ng-if="(minDate || maxDate) && (minDate != '' || maxDate != '')" class="help-block row" translate>
```
OzzieOrca: I think each element containing text added here needs to have the `translate` attribute added to it. I think it needs to be removed from the parent div and added to each of the child elements since they are not shown or hidden all at once. I'm also not sure `translate` works with html tags in it. https://angular-gettext.rocketeer.be/dev-guide/annotate/

app/views/components/pickadate.html
```diff
@@ -4,3 +4,8 @@
...
+   <label class="valid-range-msg">Valid Range:</label>
+   <span class="margin-left-10" ng-if="minDate && minDate!=''">Min Date- {{minDate | dateRangeFormat}} </span>
```
OzzieOrca: I think it would be nice for it to read `Valid Range: Aug 24, 2016 - Sep 23, 2016`. I know that causes a problem when only one end of the range is set.  Maybe you could only show the `Min` or `Max` text if the other end is not set. Something like `Valid Range: Aug 24, 2016 or later`. Even that doesn't sound quite right. Idk what is best.

If you end up leaving it how it is I think it needs a space before the dash (`Min Date - Sep 23, 2016`). Also, I don't think you need the `Date` text here as the string that follows it is clearly a date.

@hlbraddock Do we even need to have this text? I think the datepicker makes it pretty clear you can't select other dates but seeing it before the datepicker modal is opened might be useful:
screen shot 2016-09-29 at 2 21 32 pm

app/views/components/pickadate.html
```diff
@@ -4,3 +4,8 @@
   </span>
   <input type="text" placeholder="Select a date..." class="form-control" ng-disabled="disabled">
 </div>
+<div ng-if="(minDate || maxDate) && (minDate != '' || maxDate != '')" class="help-block row" translate>
+   <label class="valid-range-msg">Valid Range:</label>
+   <span class="margin-left-10" ng-if="minDate && minDate!=''">Min Date- {{minDate | dateRangeFormat}} </span>   
+   <span class="margin-left-10" ng-if="maxDate && maxDate !=''"> Max Date- {{maxDate | dateRangeFormat}}</span>
```
OzzieOrca: I don't think you need the ` && maxDate !=''` here and a couple other places. An empty string should evaluate to false.

app/scripts/directives/pickadate.js
```diff
@@ -34,12 +36,56 @@ angular.module('confRegistrationWebApp')
...
+            scope.picker.set('max',dateArray);
+          }
+        };
+		
+        //set min and max date
+        scope.setMinDate();
+        scope.setMaxDate();
+
+        scope.$watch('minDate', function (newMinDate, oldMinDate) {
+          if (angular.isUndefined(newMinDate) || newMinDate === '' || angular.equals(newMinDate, oldMinDate)) {
```
OzzieOrca: What case is the `angular.equals(newMinDate, oldMinDate)` handling? I would think that would never be true as `$watch` shouldn't call the function if there are no changes...

jayakrishnanv: @OzzieOrca  we will remove this piece of code.

app/scripts/directives/pickadate.js
```diff
@@ -34,12 +36,56 @@ angular.module('confRegistrationWebApp')
...
+          
+          var dateArray = newMaxDate.split('-');          
+          dateArray[1] = dateArray[1] - 1;         
+          scope.picker.set('max',dateArray); 
+        }, true);
+
         //when date is chosen, update model
-        var onSet = function(){
-          scope.$apply(function(){
+        var onSet = function(){
```
jayakrishnanv: @OzzieOrca  It is not needed, we will remove it.

app/views/components/blockEditor.html
```diff
@@ -95,6 +95,16 @@
                  class="form-control field-label"
                  ng-model="block.exportFieldTitle" />
         </div>
+		<div class="row form-group" ng-if="block.type === 'dateQuestion'">
+          <label class="inline-label date-alignment pull-left" translate>Min Date</label>
+          <div class="col-md-3 stacked-spacing-col-sm">
```
jayakrishnanv: we will correct it

app/views/components/blockEditor.html
```diff
@@ -95,6 +95,16 @@
...
+            <pick-a-date show-errors ng-model="block.content.range.min" ng-required="block.required" picker-max-date="block.content.range.max"></pick-a-date>
+          </div>
+          <label class="inline-label date-alignment pull-left col-md-offset-1" translate>Max Date</label>
+          <div class="col-md-3">
```
jayakrishnanv: @OzzieOrca We will correct it

app/views/components/pickadate.html
jayakrishnanv: @OzzieOrca we will add the translate attribute to each element

app/views/components/pickadate.html
jayakrishnanv: @OzzieOrca  we have changed the message to 

> Valid Range: Aug 24, 2016 - Sep 23, 2016

app/views/components/pickadate.html
jayakrishnanv: @OzzieOrca  We will correct this

jayakrishnanv:
@OzzieOrca  We have completed the changes mentioned above.

app/less/components.less
```diff
@@ -92,3 +92,18 @@ pick-a-date, pick-a-time {
...
   }
 }
+
+/****** min and max date styles *****/
+.date-alignment{
+    padding-top: 6px;
+}
+.valid-range-msg{
+ min-width:97px;
+ float:left;
```
OzzieOrca: I don't think this needs to be floating.

app/less/components.less
```diff
@@ -92,3 +92,18 @@ pick-a-date, pick-a-time {
...
+ font-weight:normal;
+ padding-left:15px;
```
OzzieOrca: You don't need this if you put the content in a Bootstrap column as mentioned below.

app/less/components.less
```diff
@@ -92,3 +92,18 @@ pick-a-date, pick-a-time {
...
+
+}
+.margin-left-10{
```
OzzieOrca: This isn't used anywhere now.

app/views/components/pickadate.html
```diff
@@ -4,3 +4,11 @@
   </span>
   <input type="text" placeholder="Select a date..." class="form-control" ng-disabled="disabled">
 </div>
+<div ng-if="minDate || maxDate" class="help-block row">
```
OzzieOrca: Since you are using `row` here, the content needs to be in a column. Probably `col-sm-12` in this case. See http://getbootstrap.com/css/#grid

app/views/components/pickadate.html
```diff
@@ -4,3 +4,11 @@
...
+   <label class="valid-range-msg" translate>Valid Range:</label>
```
OzzieOrca: Does this need to be in a label? It is styled the same as the rest of the text and you had to add `font-weight:normal;` to make it look the same.

app/less/components.less
```diff
@@ -92,3 +92,18 @@ pick-a-date, pick-a-time {
     }
   }
 }
+
+/****** min and max date styles *****/
+.date-alignment{
+    padding-top: 6px;
+}
+.valid-range-msg{
+ min-width:97px;
```
OzzieOrca: Is this needed? I guess you were trying to make the text wrap better. Is this still an issue with the shorter message?

app/views/components/pickadate.html
jayakrishnanv: @OzzieOrca Changed the style to match the rest of the text

app/views/components/pickadate.html
jayakrishnanv: @OzzieOrca moved the contents to col-sm-12

app/less/components.less
jayakrishnanv: @OzzieOrca Removed .valid-range-msg.

app/less/components.less
jayakrishnanv: @OzzieOrca Removed .valid-range-msg.

jayakrishnanv:
Tagging @jcsaplar as per Lee's suggestion

jayakrishnanv:
@OzzieOrca @hlbraddock @jcsaplar Merged this branch to #496

OzzieOrca:
Assuming this has been tested in staging, this PR is ready to merge into master. @hlbraddock @jcsaplar

PR #488, opened 2016-09-23
Event 379 valid option range for number questions

jayakrishnanv:
@hlbraddock , Implemented Min / Max range for Number Questions, testing yet to be done. Please review.

PR #497, opened 2016-09-30
Event-379 - Implemented min and max validation for number question

jayakrishnanv:
@hlbraddock New branch for min and max validation for number question

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -95,6 +95,7 @@
   <script src="scripts/directives/selectOnClick.js"></script>
   <script src="scripts/directives/showAnswer.js"></script>
   <script src="scripts/directives/showErrors.js"></script>
+  <script src="scripts/directives/stringToNumber.js"></script>
 
   <!--Filters-->
   <script src="scripts/filters/evtDateFormat.js"></script>
@@ -121,6 +122,7 @@
   <script src="scripts/services/unauthorizedInterceptor.js"></script>
   <script src="scripts/services/uuid.js"></script>
   <script src="scripts/services/validateRegistrant.js"></script>
+  <script src="scripts/services/util.js"></script>
   <!-- endbuild -->
   <!-- build:js scripts/errorNotify.js -->
   <script src="scripts/errorNotify.js"></script>
--- a/app/less/components.less
+++ b/app/less/components.less
@@ -92,3 +92,29 @@ pick-a-date, pick-a-time {
     }
   }
 }
+
+
+/* componet styles */
+.valid-range-msg{
+ min-width:97px;
+ float:left;
+ font-weight:normal;
+ padding-left:15px;
+}
+.margin-left-10{
+	margin-left:10px;
+}
+
+.invalid-range-input{
+    border-color: #a94442;
+    -webkit-box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075);
+    box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075);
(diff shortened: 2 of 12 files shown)


OzzieOrca:
What's the difference between this and https://github.com/CruGlobal/conf-registration-web/pull/488?

jayakrishnanv:
@OzzieOrca This branch is a fresh branch which contains only the code changes related to EVENT-379. Please discard the old branch https://github.com/CruGlobal/conf-registration-web/pull/488 and continue testing using this new branch.

jayakrishnanv:
@OzzieOrca Please understand that we have moved the min / max validation to the "Close"button click, as it is not practical to validate these on each key stroke. To achieve this, we have bound the min / max text boxes to a temporary model, instead of the default "block" model, so that the values do not get submitted to the back-end on each key stroke. Instead. the values are bound to a temporary model, and on clicking the "Close" button of the Edit pane, the values are validated and if found valid, are bound to the "block" model and sent to back-end, otherwise a validation error message is displayed. Please base your testing and comments based on this logic.

jayakrishnanv:
Tagging @jcsaplar as per Lee's suggestion

jcsaplar:
This looks good on the front end but I have a question in regards to the error message "Please enter a valid number." Will this change based on taking out the original message out when there is no default value present? @jayakrishnanv

jayakrishnanv:
@jscaplar We can remove this message and still show the message if the default value falls outside the min / max range

OzzieOrca:
@jayakrishnanv Can you try implementing this with Angular's `ng-min` and `ng-max directives? If you've tried this already and ran into some issue, can you let me know what that was so I can help?
https://docs.angularjs.org/api/ng/input/input%5Bnumber%5D I can work with you on it if you're not sure how do do it.

Angular should be able to validate the input on each keystroke, but if that is an issue for some reason we still should be able to use a temp value to edit the ranges and then pass that to the block's input on close. Hopefully that won't be needed though. I'm not sure if that was just getting too complicated for your custom code or if for some reason we don't want to validate the default value until the block edit panel is closed.

jayakrishnanv:
@OzzieOrca 

  We tried implementing this with Angular's ng-min and ng-max directives, but faced some issue. Even though the slider arrows restrict the values to be within the min / max range, the user can always key in a value outside the range, in the number box. Although this triggers a validation error, the entered value stays in the number box and in the back-end, the values is sent as empty (undefined) and hence the value is updated as blank in the database. So, even though the values is blank in the database, the user sees the entered invalid value in the number box.

Apart from this, we can't rely on the validation triggered on every key stroke because, the user may not have completed entering the value in the number box. For example, if the user enters 10 in the min box and tries to enter 100 in the max box, it would trigger a validation error on entering the very first digit, i.e., 1, as 1 is less than 10, which is invalid. So, we resorted to binding the number box to a temporary model, than the default "block" model, so that no validation error is triggered on key stroke, and nor is the value sent to db. The entered value is validated only on clicking the "close" button and if the validation fails, the panel stays open with the error messages displayed below the number box. If the validation succeeds, the value is bound to the "block" model and is sent to the database. The same goes for checking whether the default value entered in the number box lies within the min / max range.
	
This logic, we resorted to, seems to be working fine. Do you want us to revert to using the Angular's ng-min and ng-max directives, instead?

hlbraddock:
Closed in favor of replacement https://github.com/CruGlobal/conf-registration-web/pull/499

PR #492, opened 2016-09-27
Event 432 rules support for any match

jayakrishnanv:
@hlbraddock , Implemented option for choosing either "OR" or "AND" as the Rule Operand. Please test.

PR #499, opened 2016-10-05
EVENT-379: Enable min/max validation on numberQuestion. Fix numberQuestion rules

OzzieOrca:
This PR replaces https://github.com/CruGlobal/conf-registration-web/pull/497

- Add min and max inputs to Advanced tab of block editor
- Validate number input in rules section but allow invalid so updates to the number validation range don't cause api errors. The user can update it later.
- Add stringToNumber directive and casting to blockVisibleRuleCheck since rules are stored as strings

@jayakrishnanv What do you think of this? I was trying to make sure it worked and show you an example but I just ended up implementing it all. I started with your code but got rid of all the custom validation and DOM manipulation. The validation wasn't too bad but I spent a while trying to fix the rules so they would compare numbers as numbers instead of strings which I assume is a long standing bug. Let me know if you see any cases where this doesn't work or if you have any improvements.

It might need some code to interact with the default value feature correctly

@hlbraddock How easy would it be to convert `rule.value` to a number in the api? I got it to work  as a string but it is more code than it needs to be. If it's not trivial, don't worry about it.

### DIFF
--- a/app/scripts/services/validateRegistrant.js
+++ b/app/scripts/services/validateRegistrant.js
@@ -12,13 +12,16 @@ angular.module('confRegistrationWebApp')
         var rule = block.rules[i];
         var answer = _.find(answers, { blockId: rule.parentBlockId });
         if (angular.isDefined(answer) && answer.value !== '') {
-          if (rule.operator === '=' && answer.value === rule.value) {
+          //If string is a number, parse it as a float for numerical comparison
+          var answerValue = !isNaN(answer.value) ? parseFloat(answer.value) : answer.value;
+          var ruleValue = !isNaN(rule.value) ? parseFloat(rule.value) : rule.value;
+          if (rule.operator === '=' && answerValue === ruleValue) {
             validRuleCount++;
-          } else if (rule.operator === '!=' && answer.value !== rule.value) {
+          } else if (rule.operator === '!=' && answerValue !== ruleValue) {
             validRuleCount++;
-          } else if (rule.operator === '>' && answer.value > rule.value) {
+          } else if (rule.operator === '>' && answerValue > ruleValue) {
             validRuleCount++;
-          } else if (rule.operator === '<' && answer.value < rule.value) {
+          } else if (rule.operator === '<' && answerValue < ruleValue) {
             validRuleCount++;
           }
         }
@@ -27,13 +30,9 @@ angular.module('confRegistrationWebApp')
         }
       }
 
-      if (block.rules.length === 0 || // If no rules are set
+      return block.rules.length === 0 || // If no rules are set
         (ruleOperand === 'OR' && validRuleCount > 0) ||
-        (ruleOperand === 'AND' && validRuleCount === block.rules.length)) {
-        return true;
-      } else {
-        return false;
-      }
+        (ruleOperand === 'AND' && validRuleCount === block.rules.length);
     };
 
     var blockInRegistrantType = function(block, registrant){
--- a/app/index.html
+++ b/app/index.html
@@ -95,6 +95,7 @@
   <script src="scripts/directives/selectOnClick.js"></script>
   <script src="scripts/directives/showAnswer.js"></script>
   <script src="scripts/directives/showErrors.js"></script>
+  <script src="scripts/directives/stringToNumber.js"></script>
 
   <!--Filters-->
   <script src="scripts/filters/evtDateFormat.js"></script>
--- a/app/scripts/directives/blockEditor.js
+++ b/app/scripts/directives/blockEditor.js
@@ -218,6 +218,8 @@ angular.module('confRegistrationWebApp')
               return ['M', 'F'];
             case 'yearInSchoolQuestion':
               return ['Freshman', 'Sophomore', 'Junior', 'Senior', 'Graduate Student'];
+            case 'numberQuestion':
+              return block.content.range;
             default:
               return [];
           }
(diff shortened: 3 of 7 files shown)


app/scripts/services/validateRegistrant.js
```diff
@@ -8,16 +8,19 @@ angular.module('confRegistrationWebApp')
...
       angular.forEach(block.rules, function(rule){
         var answer = _.find(answers, {blockId: rule.parentBlockId});
-        if(angular.isUndefined(answer) || answer.value === ''){
+        if(angular.isUndefined(answer) || answer === null || answer.value === ''){
           returnValue = false;
         }else{
-          if(rule.operator === '=' && answer.value !== rule.value) {
+          //If string is a number, parse it as a float for numerical comparison
+          var answerValue = !isNaN(answer.value) ? parseFloat(answer.value) : answer.value;
+          var ruleValue = !isNaN(rule.value) ? parseFloat(rule.value) : rule.value;
```
jayakrishnanv: @OzzieOrca I think you should have tried to fix the existing bug in rules logic in a new separate branch or at least in #492 . As you said, the type of question for the rules is not available in this context.

jayakrishnanv: @OzzieOrca It seems like there are no issues with this version, can you merge this to master, then?

OzzieOrca: Ya, you're probably right. Oh well. I could split it out if this doesn't get merged immediately.

OzzieOrca:
@jayakrishnanv are you ok with me closing https://github.com/CruGlobal/conf-registration-web/pull/497 in favor of this PR. I feel bad tossing some of your code but I think using the angular validation makes the code much cleaner and more maintainable in the future. It was helpful to have some of the changes already made and know where to write code/add form validation. Thanks :)

Also do you feel comfortable reviewing and approving this PR or would you rather I ask one of the other front end guys at Cru?

jayakrishnanv:
@OzzieOrca We tested this version and everything seems fine except a couple of things:

1. In the rules section, if the user adds a number question with min / max range set, and enter a value which is out of the min / max range, then despite displaying a validation error, this invalid value is saved in the database.

2. The validation message "Valid Range: 10 or higher", etc., is absent for number questions with min / max range set.

hlbraddock:
@OzzieOrca Regarding converting `rule.value` to a number in the api, it's doable. But, as there are currently no front end to back end tightly coupled dependences requiring synchronized deployment, I'd like not to add one in this first release if we can avoid it. However, I'm all for changing it once the dust has settled, there are less moving parts, and we can make that one change. Acceptable?

OzzieOrca:
@jayakrishnanv What should happen if the range is updated and makes the rule value invalid? Should we clear it? Should we keep the old value locally until a page refresh? Having an invalid value there will just make the rule never trigger which I don't think is that bad. No matter what we do the user still has to update it for it to work right again. We could try to make that error more visible to the user. I chose to `allowInvalid` because the api throws an error if the rule value is set to null which is exactly what ng-model does when the input is invalid. It doesn't seem to care that we are storing an out of range value there. I could try to get the rule input to save an empty string or something the api is ok with storing when it is invalid. But if the user refreshes the page they'll lose their old value which probably isn't a big deal but idk.

@hlbraddock Do we want the number range to always be shown to the user? Currently it looks like this:
screen shot 2016-10-06 at 8 52 18 am
The user has to type something to see the message:
screen shot 2016-10-06 at 8 52 28 am

@hlbraddock That's fine. I mentioned it because it seems like the front end had always been assuming that number was a number and not a string. The greater than/less than rules seemed to be comparing strings which just compares them alphabetically...

hlbraddock:
@jayakrishnanv I think it's ok to show the number range after the user types.

hlbraddock:
@OzzieOrca We of course don't want number strings compared alphabetically but rather by converting number and comparing. I presume you did so herein?

jayakrishnanv:
@OzzieOrca Scotty,
  As you said, this is not a major issue, as an error message is displayed when the rule value is out of range and the administrator is most likely to update it to a valid number.

  If you really want to tackle this issue, you should not let the invalid value to be saved on the db. This is one of the reasons why we resorted to validate / save on close button click. If you feel this should not be an issue, you can go ahead with this code.

OzzieOrca:
@hlbraddock Yes this converts strings to numbers before comparing them. If the api returned a string for number questions we wouldn't need that front end code. It should be good to go as is though. We can revisit it later if desired.

@jayakrishnanv I think it's fine. The now invalid value is already saved there. I think if you did the validation on close, you'd still have the same problem. If you want to kill the old value and save null to the server, you can get rid of ng-model-options="{ allowInvalid: true }" and write some directive to or add code to the code that saves this value to the server and set it to null or something instead of undefined. Right now it isn't sending that key at all if the input is undefined.

jayakrishnanv:
@OzzieOrca @hlbraddock Just wanted to let you know that we came across a small issue in rules section related to min / max number validation.

  If the first question type in the rules drop down box is a number question with min / max range set, then an out of range validation error is displayed even if we choose another question from the drop down box.

 This maybe because you have set the "form-group"class to the outer div, moving it to the inner div for Number questions should solve the issue:

```html
<div class="col-md-5 col-sm-7 stacked-spacing-col-md form-group" ng-switch="ruleValueInputType(rule.parentBlockId)">
              <div ng-switch-when="number">
                <input type="number"
```

hlbraddock:
@jayakrishnanv If you have a bug fix, create a new branch, and submit a PR.

jayakrishnanv:
@hlbraddock @OzzieOrca This is a bug which we identified in Scotty's version of EVENT-379. I guess this can be fixed by him.

OzzieOrca:
@jayakrishnanv Good catch. Thanks for pointing that out. I made a PR to fix it.

PR #511, opened 2016-10-14
Event 456 set help email

jcsaplar:
Changed email addresses to help@cru.org from apps@cru.org


jcsaplar:
@hlbraddock please confirm this is what you were referencing in the JIRA task

PR #512, opened 2016-10-14
In validateRegistrant, handle case where rules may be missing. Show b…

OzzieOrca:
…lock.content for info question if block.content.paragraph is undefined.

### DIFF
--- a/app/views/components/blockRegistration.html
+++ b/app/views/components/blockRegistration.html
@@ -2,7 +2,8 @@
   <label class="control-label"><span ng-bind-html="block.title"></span><em class="required" ng-if="block.required">*</em></label>
 
   <ng-switch on="block.type">
-    <p ng-switch-when="paragraphContent" ng-bind-html="block.content.paragraph" class="display-linebreaks"></p>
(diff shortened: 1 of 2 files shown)


app/views/components/blockRegistration.html
```diff
@@ -2,7 +2,8 @@
   <label class="control-label"><span ng-bind-html="block.title"></span><em class="required" ng-if="block.required">*</em></label>
 
   <ng-switch on="block.type">
-    <p ng-switch-when="paragraphContent" ng-bind-html="block.content.paragraph" class="display-linebreaks"></p>
+    <p ng-switch-when="paragraphContent" ng-if="block.content.paragraph" ng-bind-html="block.content.paragraph" class="display-linebreaks"></p>
+    <p ng-switch-when="paragraphContent" ng-if="block.content.paragraph === undefined" ng-bind-html="block.content" class="display-linebreaks"></p>
```
OzzieOrca: I may need to check if `block.content` is a string instead. Idk what `block.content.paragraph` gets set to if the user leaves it blank. I think I'd have to add a function to the scope to do that. Worst case is it show the user [Object object] if the admin added a paragraph question and did not fill it with any content.

PR #514, opened 2016-10-17
Change blockVisibleRuleCheck to use forEach which handles the case wh…

OzzieOrca:
…ere block.rules may not be defined

### DIFF
--- a/app/scripts/services/validateRegistrant.js
+++ b/app/scripts/services/validateRegistrant.js
@@ -1,15 +1,18 @@
 'use strict';
 
 angular.module('confRegistrationWebApp')
-  .service('validateRegistrant', function validateRegistrant() {
+  .service('validateRegistrant', function validateRegistrant($window) {
 
     var blockVisibleRuleCheck = function(block, registrant){
       var answers = registrant.answers;
       var ruleOperand = block.content && block.content.ruleoperand ? block.content.ruleoperand : 'AND';
       var validRuleCount = 0;
 
-      for (var i = 0; i < block.rules.length; i++) {
-        var rule = block.rules[i];
+      if($window.Rollbar && !_.isArray(block.rules)){
+        $window.Rollbar.info('Block rules value in blockVisibleRuleCheck, typeof: ' + typeof(block.rules) + ', JSON.stringify: ' + JSON.stringify(block.rules));
+      }
+
+      _.forEach(block.rules, function(rule, i){
         var answer = _.find(answers, { blockId: rule.parentBlockId });
         if (angular.isDefined(answer) && answer.value !== '') {
           //If string is a number, parse it as a float for numerical comparison
@@ -26,9 +29,9 @@ angular.module('confRegistrationWebApp')
           }
         }
         if ((ruleOperand === 'OR' && validRuleCount > 0) || (ruleOperand === 'AND' && validRuleCount <= i)) {
-          break;
+          return false; // Exit lodash foreach as we found a case which determines the whole outcome of this function
         }
-      }
+      });
 
       return !block.rules || block.rules.length === 0 || // If no rules are set
         (ruleOperand === 'OR' && validRuleCount > 0) ||


app/scripts/services/validateRegistrant.js
```diff
@@ -26,9 +25,9 @@ angular.module('confRegistrationWebApp')
           }
         }
         if ((ruleOperand === 'OR' && validRuleCount > 0) || (ruleOperand === 'AND' && validRuleCount <= i)) {
-          break;
+          return false; // Exit lodash foreach as we found a case which determines the whole outcome of this function
```
hlbraddock: You wouldn't want to return `false` if `(ruleOperand === 'OR' && validRuleCount > 0)`.

Maybe instead:
```
if (ruleOperand === 'AND' && validRuleCount <= i) {
      return false; // Exit lodash foreach as we found a case which determines the whole outcome of this function
}
if (ruleOperand === 'OR' && validRuleCount > 0) {
      return true; // Exit lodash foreach as we found a case which determines the whole outcome of this function   
}
```

app/scripts/services/validateRegistrant.js
```diff
@@ -26,9 +25,9 @@ angular.module('confRegistrationWebApp')
...
         }
-      }
+      });
 
       return !block.rules || block.rules.length === 0 || // If no rules are set
```
hlbraddock: I presume that if  `!block.rules` is true then `**block.rules.length === 0` does not get evaluated. @OzzieOrca

app/scripts/services/validateRegistrant.js
OzzieOrca: `return false;` (`return;` would probably work too) exits the for loop (which is a function) early and not the entire function. Lodash wraps the iterations in a function.

I did have a version that I suggested to the contractors where you returned values right there but they didn't like it. I was trying to change as few things as possible to reduce merge conflicts.

You did just make me notice that it is relying on `i` in `validRuleCount <= i`. So don't merge it. Let me fix that. Good catch.

app/scripts/services/validateRegistrant.js
hlbraddock: @OzzieOrca 

With the conditionals above, I would think that
```
      return !block.rules || block.rules.length === 0 || // If no rules are set
        (ruleOperand === 'OR' && validRuleCount > 0) ||
        (ruleOperand === 'AND' && validRuleCount === block.rules.length);
```

could be reduced to
```
      return !block.rules || block.rules.length === 0; // If no rules are set;
```

app/scripts/services/validateRegistrant.js
```diff
@@ -8,8 +8,9 @@ angular.module('confRegistrationWebApp')
       var ruleOperand = block.content && block.content.ruleoperand ? block.content.ruleoperand : 'AND';
       var validRuleCount = 0;
 
-      for (var i = 0; i < block.rules.length; i++) {
-        var rule = block.rules[i];
+      Rollbar && Rollbar.info('Block rules value in blockVisibleRuleCheck' + JSON.stringify(block.rules));
```
hlbraddock: @OzzieOrca 
Should we limit the log to Rollbar only if `!block.rules` (that is, if `block.rules` is not defined)?

hlbraddock:
@OzzieOrca 
https://travis-ci.org/CruGlobal/conf-registration-web failed 

```
     12 |        Rollbar && Rollbar.info('Block rules value in blockVisibleRuleCheck, typeof: ' + typeof(block.rules) + ', JSON.stringify: ' + JSON.stringify(block.rules));
                                                                                                                                                                          ^ Expected an assignment or function call and instead saw an expression.
     12 |        Rollbar && Rollbar.info('Block rules value in blockVisibleRuleCheck, typeof: ' + typeof(block.rules) + ', JSON.stringify: ' + JSON.stringify(block.rules));
                 ^ 'Rollbar' is not defined.
     12 |        Rollbar && Rollbar.info('Block rules value in blockVisibleRuleCheck, typeof: ' + typeof(block.rules) + ', JSON.stringify: ' + JSON.stringify(block.rules));
                            ^ 'Rollbar' is not defined.
>> 3 errors in 71 files
Warning: Task "jshint:all" failed. Use --force to continue.
Aborted due to warnings.
```

OzzieOrca:
Fun. I'll have to tell JSHint to ignore it.

PR #515, opened 2016-10-19
Pass ruleType to blockVisibleRuleCheck in validate

OzzieOrca:
- Add rollbar error if ruleType is ever not defined

### DIFF
--- a/app/scripts/services/validateRegistrant.js
+++ b/app/scripts/services/validateRegistrant.js
@@ -16,6 +16,9 @@ angular.module('confRegistrationWebApp')
         blockTypeSpecificRules = _.filter(block.rules, { 'ruleType': ruleType });
         ruleOperand = block.content && block.content.forceSelectionRuleOperand ? block.content.forceSelectionRuleOperand : 'AND';
       } else {
+        if($window.Rollbar){
+          $window.Rollbar.error('blockVisibleRuleCheck was called with an unknown rule type: ', ruleType);
+        }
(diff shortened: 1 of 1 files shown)


app/scripts/services/validateRegistrant.js
```diff
@@ -68,7 +71,7 @@ angular.module('confRegistrationWebApp')
       var blocks = page ? _.find(conference.registrationPages, {id: page}).blocks : _.flatten(conference.registrationPages, 'blocks');
 
       angular.forEach(blocks, function(block){
-        if (!block.required || block.adminOnly || !blockVisibleRuleCheck(block, registrant) || !blockInRegistrantType(block, registrant)) { return; }
+        if (!block.required || block.adminOnly || !blockVisibleRuleCheck(block, registrant, ruleTypeConstants.SHOW_QUESTION) || !blockInRegistrantType(block, registrant)) { return; }
```
hlbraddock: Does Angular not require you to pass all expected parameters? @OzzieOrca

OzzieOrca: Javascript isn't very strict. Anything you don't pass results in undefined (which is similar but very different to `null`). If you want checking like that I think you'd need to migrate to Typescript (which is recommended for Angular 2) or something similar.

I think this is one of my favorite images:
Javascript the good parts

hlbraddock: Haha! :)

PR #516, opened 2016-11-11
Event-461 - Implemented rules based on checkbox questions

jayakrishnanv:
@hlbraddock, @jcsaplar @OzzieOrca  Implemented rules based on checkbox questions, as per https://jira.cru.org/browse/EVENT-461. Please review.

### DIFF
--- a/app/scripts/directives/blockEditor.js
+++ b/app/scripts/directives/blockEditor.js
@@ -11,7 +11,7 @@ angular.module('confRegistrationWebApp')
         $scope.showClearBtn = true;
         $scope.isAdmin = true;
         $scope.ruleTypeConstants = ruleTypeConstants;
-        
+
         $scope.popup = {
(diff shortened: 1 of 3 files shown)


hlbraddock:
@OzzieOrca Can you review this relatively minor addition, as per https://jira.cru.org/browse/EVENT-461.

app/scripts/directives/blockEditor.js
```diff
@@ -11,7 +11,7 @@ angular.module('confRegistrationWebApp')
         $scope.showClearBtn = true;
         $scope.isAdmin = true;
         $scope.ruleTypeConstants = ruleTypeConstants;
-
```
OzzieOrca: This seems like it will never get run as the `$scope.block.type === 'paragraphContent'` condition above would make sure the type is not `checkboxQuestion`.

PR #518, opened 2016-12-14
Add DTM code.

twinge:


twinge:
@adammeyer My goal was to have a different <script> include on prod vs staging. Could you please confirm that I did that right? I was just guessing based on other code in the file.

adammeyer:
I think you need to swap them.  Currently it's set to remove the stage script on stage.  I believe you want to remove the prod script on stage and the stage script on dist.

PR #520, opened 2017-01-13
Upgrade to jQuery 3

canac:
Upgrade to jQuery 3.0.0 and compensate for any breaking changes.


canac:
It works for me. I modified `package.json`. Did you run `npm install`?

OzzieOrca:
I thought that was going to be it but I just ran npm i and bower install... I still have that TNT error and now get moment.isDate is not a function and the datepickers still don't work... I could be doing something wrong...

canac:
Try this now. Note that the date picker is using a deprecated moment API that results in a warning being printed to the console. There is an open issue about that: Eonasdan/bootstrap-datetimepicker#1773. I don't know of anything we can do to get rid of that warning besides waiting for a fix.

PR #521, opened 2017-01-13
Modify the payment options form for TSYS

canac:
Implement [EVENT-484](https://jira.cru.org/browse/EVENT-484), modifying fields and labels in the credit card payment options form. @hlbraddock please make sure that my changes match your expectations.

This PR will break things if merged into master. That is why I am requesting to merge it into the EVENT-485 branch, which can collect PRs for subtasks until this feature is completed, at which point we can merge that branch into master.

### DIFF
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -25,9 +25,21 @@ angular.module('confRegistrationWebApp')
     };
     $scope.changeTab($scope.tabs[0]);
 
-    $scope.paymentGateways = [
-      {id: 'AUTHORIZE_NET', name: 'Authorize.Net'}
-    ];
+    $scope.paymentGateways = {
+      AUTHORIZE_NET: {
+        name: 'Authorize.Net',
+        fields: {
+          paymentGatewayId: { title: 'Account ID' },
+          paymentGatewayKey: { title: 'Key' }
+        }
+      },
+      TSYS: {
+        name: 'TSYS',
+        fields: {
+          paymentGatewayId: { title: 'Merchant Device' }
+        }
+      }
+    };
 
     $scope.conference = angular.copy(conference);
 
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -208,7 +208,9 @@ angular.module('confRegistrationWebApp')
 
       //Credit cards
       if (_.isEmpty($scope.conference.paymentGatewayId) && _.some($scope.conference.registrantTypes, 'acceptCreditCards')) {
-        validationErrors.push('Please enter a credit card Account ID and Key under the "Payment Options" tab.');
+        var paymentGateway = $scope.paymentGateways[$scope.conference.paymentGatewayType];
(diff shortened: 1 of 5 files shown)


app/scripts/controllers/eventDetails.js
```diff
@@ -25,10 +25,6 @@ angular.module('confRegistrationWebApp')
     };
     $scope.changeTab($scope.tabs[0]);
 
-    $scope.paymentGateways = [
```
hlbraddock: Isn't `paymentGateways` used in some context in this app?

hlbraddock:
From a UI standpoint, these changes look right. +1

app/scripts/controllers/eventDetails.js
canac: It is only used to generate the "Payment Gateway" selector that I removed in this PR.

canac:
@OzzieOrca Can you give this a quick technical review?

canac:
@hlbraddock What is your opinion on having TSYS (or TCEP?) in our labels vs. a more generic phrase?

hlbraddock:
We had listed `Authorize.NET` as a payment gateway option (the only option we afforded) since you could theoretically go out and get your own Authnet id and key and use them at your leisure.  Therefore, you needed to know the actual payment gateway provider.

However, in this new integration, you don't have the option of selecting a payment gateway. From that standpoint, the conference administrator doesn't need to know anything about the backend payment provider. 

For this reason, I thought therefore to make the solution and verbiage more generic in nature, limiting it to 'merchant id' or 'merchant device id' (the latter being slightly more accurate).

That said, you may want to cleanup some references to `Authorize.NET`. `paymentGatewayType` is currently being ignored (though ideally it should removed altogether from the API, I've not yet done that sort of cleanup on the backend).

canac:
Updated after a discussion with @hlbraddock. 

Now newly-created conferences will be configured to use TSYS. Existing conferences will continue using Authorize.NET. Conferences' credit card processing provider cannot be changed. A short message is added to the credit card form that indicates which provider (TSYS or Authorize.NET) the conference is using.

@OzzieOrca and @hlbraddock Please re-review.

app/scripts/controllers/eventDetails.js
```diff
@@ -25,9 +25,21 @@ angular.module('confRegistrationWebApp')
...
+        name: 'Authorize.Net',
+        fields: {
+          paymentGatewayId: { title: 'Account ID' },
+          paymentGatewayKey: { title: 'Key' }
+        }
+      },
+      TSYS: {
+        name: 'TSYS',
+        fields: {
+          paymentGatewayId: { title: 'Merchant Device' }
```
hlbraddock: We actually should probably not name it `merchant` since there exists a `merchant id` as distinct from a `device id`. I had thought `merchant` was more UI friendly, but it's actually not accurate and could be confusing as well as potentially wrong if an actual `merchant id` is provided. 

For now, let's rename this `ERT Device ID`.

@canac

canac: ✔️  Fixed and ready for re-review.

PR #522, opened 2017-01-23
Update moment-timezone dependency

canac:
Update the `moment-timezone` dependency to remove deprecation warnings about `moment().zone`.


canac:
I'm going to postpone the moment upgrade because the purpose of this change is to remove that deprecation warning, not to upgrade all of our dependencies (although that would be good to do at some point).

PR #523, opened 2017-01-26
Support payments through TSYS

canac:
Tokenize credit card payments via TSYS using the `cru-payments` library.

### DIFF
--- a/app/scripts/directives/payment.js
+++ b/app/scripts/directives/payment.js
@@ -138,12 +138,13 @@ angular.module('confRegistrationWebApp')
 
                 // Run all of the validations
                 _.forEach(validations, function (validators, field) {
-                  // Validate this field
-                  validators.forEach(function (validator) {
-                    if (!validator.validate(currentPayment.creditCard[field])) {
-                      paymentErrors.push(gettext(validator.errorMessage));
-                    }
-                  })
+                  // Validate this field, stopping on the first failed validator
+                  var failedValidator = _.find(validators, function (validator) {
+                    return !validator.validate(currentPayment.creditCard[field]);
+                  });
+                  if (failedValidator) {
+                    paymentErrors.push(gettext(failedValidator.errorMessage));
+                  }
                 });
 
                 //don't require billing address if admin payment
--- a/app/scripts/directives/payment.js
+++ b/app/scripts/directives/payment.js
@@ -11,7 +11,7 @@ angular.module('confRegistrationWebApp')
         paymentMethods: '=paymentMethods',
         isAdminPayment: '=adminPayment'
       },
-      controller: function ($scope, $http, cruPayments, expenseTypesConstants, gettext) {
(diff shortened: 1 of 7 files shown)


app/scripts/directives/payment.js
```diff
@@ -64,21 +64,88 @@ angular.module('confRegistrationWebApp')
...
+                    }
+                  ]
+                };
+
+                // Run all of the validations
+                _.forEach(validations, function (validators, field) {
+                  // Validate this field
+                  validators.forEach(function (validator) {
+                    if (!validator.validate(currentPayment.creditCard[field])) {
+                      paymentErrors.push(gettext(validator.errorMessage));
```
OzzieOrca: As far as I know, gettext does static translation. Which mean it goes through the app, finds these calls and prepares what is directly in them for translation. I think it will try to translate "validator.errorMessage" and won't actually execute any js to figure out what could go in there. Then at runtime, this gets replaced with the correct translation.

app/scripts/services/error.js
```diff
@@ -6,6 +6,8 @@ angular.module('confRegistrationWebApp')
       // Generate a promise catch handler that generates an Error object from an HTTP response object
       errorFromResponse: function (defaultErrorMessage) {
         return function (res) {
+          console.error(res);
```
OzzieOrca: Was this console on purpose? Just making sure it wasn't left over from development.

dougworner: @OzzieOrca Hey Scotty,  was just following your last post.  I agree with you that giving validation feedback next to each input (where possible, or maybe in the same pop up modal) is better UX.  I chimed in because I'm actually working on an Jira issue for ERT (EVENT-489) where there is supposed to be an error popup in the browser that isn't popping up because getting a 500 error from server.  Also made not in code of where in code it is specifically failing. Attached is picture.  Are these the modal with errors that you are referring too?  Because for this particular issue I think having a validation message within the paymentModal.js window would be much better than the notify error message (see attached) that is failing.
notify_error_message

canac: ✔️  Yes, that logging is intentional, but I just switched it to use `$log` since that is best-practice for Angular apps.

hlbraddock:
@OzzieOrca I'm open to incorporating your UI input feedback suggestions, but, unless they are trivial to implement, I think I'd prefer to do it in a separate PR, in order not to add any possible delay to the critical PCI deadlines. @canac

canac:
I agree with @OzzieOrca that the validation changes would be good from a UX perspective, but they would not be trivial to implement (my rough estimate is on the order of a day of work). I am with @hlbraddock on avoiding increasing the scope of this task because doing so would delay its completion.

app/scripts/directives/payment.js
canac: I researched it, and, yes. you are correct. `gettext` doesn't actually do any translation, whereas `gettextCatalog.getString` does. Both forms mark the string as needing translation. For this reason, I switched to using the latter.

PR #524, opened 2017-01-31
analytics

adammeyer:


app/scripts/controllers/landing.js
```diff
@@ -41,6 +41,11 @@ angular.module('confRegistrationWebApp')
         };
         $scope.searchComplete = val;
         $scope.searchResults = response.data;
+
+        analytics.digitalData.searchFilter = '';
+        analytics.digitalData.searchTeam = val;
```
canac: Is this supposed to be `searchTerm`?

app/scripts/factories/analytics.js
```diff
@@ -0,0 +1,27 @@
...
+      });
+
+      //page title
+      var pageTitle = $rootScope.globalPage ? $rootScope.globalPage.title : null;
+      $window.digitalData.page = {
+        pageInfo: {
+          pageName: 'Event Registration Tool | ' + (pageTitle ? pageTitle : 'no PageName found')
+        }
+      };
+      $window._satellite.pageBottom();
```
canac: Won't this throw an exception is `$window._satellite` is undefined?

app/scripts/factories/analytics.js
```diff
@@ -0,0 +1,27 @@
...
+        $window.digitalData.loggedInStatus = 'Logged In';
+        //TODO: add GR id once exposed by API
+        //$window.digitalData.GUID = '';
+      });
+
+      //page title
+      var pageTitle = $rootScope.globalPage ? $rootScope.globalPage.title : null;
+      $window.digitalData.page = {
+        pageInfo: {
+          pageName: 'Event Registration Tool | ' + (pageTitle ? pageTitle : 'no PageName found')
```
canac: Ternary expression can be simplified to `pageTitle || 'no PageName found'`

app/index.html
```diff
@@ -308,6 +303,15 @@ <h1 ng-bind="globalPage.title"></h1>
   </script>
   <!-- /build -->
 
-  <script type="text/javascript">_satellite.pageBottom();</script>
+  <!-- Adobe DTM -->
+  <script>
+    window.digitalData = {};
+  </script>
+  <!-- build:remove:dist -->
```
OzzieOrca: This is currently an issue from Josh's initial implementation but both of these scripts are available on dev when they shouldn't be. I'm not sure of a good way to fix it with [grunt-processhtml](https://www.npmjs.com/package/grunt-processhtml). Maybe we could comment them out and then remove the comments but that seems messier...

adammeyer: I tried a few different plugins this afternoon.  What is the negative of just leaving both scripts in for development?  I don't think local devs will need DTM functioning.

OzzieOrca: I'd assume we don't want to track dev pageviews as production pageviews. The staging doesn't matter much.

adammeyer: True.  Forgot about that.  Take a look at what I committed and let me know what you think.

OzzieOrca: Idk... Loading dynamically with jQuery works... It just doesn't feel like the right way to do that... If grunt can't do it, maybe we should move the ERT to webpack. I don't think it would take too much to just import a list of files and get the css working. We don't need to make it do anything fancy at first although we would need to add the translation tooling, assuming that has started being used.

Also that seems like it will delay the loading of DTM. From Adobe's DTM docs:

> Why does dynamic tag management have two parts?
> Dynamic tag management allows you to precisely time the loading of javascript and page content in your page. You can choose to run rules at the beginning or end of a page. This allows you to implement testing tools and other technologies, while retaining full control over tracking your pages.

Idk that Cru is doing anything with analytics that needs to distinguish between page top and bottom but idk what impact this will have on what's built into analytics.

Maybe this issue should have been addressed outside of this PR... It's my fault for bringing it up haha 😄  Or maybe just go with your jQuery solution and we'll hope to upgrade it soon.

app/scripts/controllers/landing.js
OzzieOrca: @adammeyer Did this get addressed? Is it correct as is or is it a spelling error?

app/index.html
adammeyer: Check out my latest commit.  I swapped out jquery and use grunt processhtml.  Thoughts?

app/index.html
```diff
@@ -308,6 +307,6 @@ <h1 ng-bind="globalPage.title"></h1>
   </script>
   <!-- /build -->
 
-  <script type="text/javascript">_satellite.pageBottom();</script>
+  <script>if(_satellite){ _satellite.pageBottom(); }</script>
```
adammeyer: Thanks, I will fix that.

app/scripts/factories/analytics.js
```diff
@@ -0,0 +1,31 @@
...
+
+      //page title
+      var pageTitle = $rootScope.globalPage ? $rootScope.globalPage.title : null;
+      $window.digitalData.page = {
+        pageInfo: {
+          pageName: 'Event Registration Tool | ' + (pageTitle || 'no PageName found')
+        }
+      };
+      if($window._satellite){
+        $window._satellite.pageBottom();
```
adammeyer: I forgot I already added the pageBottom call in the angular app.  I'll read the docs to get a better idea on where it should go. thanks!

adammeyer:
@OzzieOrca I added angulartics, let me know your thoughts.  Thanks!

adammeyer:
@OzzieOrca Also, if you have any input on the karma failures, let me know. Thanks!

OzzieOrca:
Closed by https://github.com/CruGlobal/conf-registration-web/pull/544

PR #526, opened 2017-02-07
Use static TSYS device ID

canac:
Use a single, hard coded device ID to interact with all TSEP services. The front end passes that device ID to the back end in exchange for a manifest. The front end also collects a merchant ID for each conference, instead of a device ID.


app/scripts/app.js
```diff
@@ -280,6 +280,32 @@ angular.module('confRegistrationWebApp', ['ngRoute', 'ngCookies', 'ngSanitize',
     $httpProvider.interceptors.push('unauthorizedInterceptor');
     $httpProvider.interceptors.push('statusInterceptor');
   })
+  .config(function (envServiceProvider) {
+    envServiceProvider.config({
+      domains: {
+        development: ['localhost'],
+        staging: ['stage.eventregistrationtool.com'],
+        production: ['stage.eventregistrationtool.com']
```
OzzieOrca: This is wrong.

canac: ✔️  Good catch! 👍

PR #528, opened 2017-02-14
Revert spouse registration

canac:
This reverts all changes specifically related to spouse registration. It keeps the code refactorings that were made while implementing spouse registration merging. [This diff](https://github.com/CruGlobal/conf-registration-web/compare/88d3f2f...d889ba9#diff-0ae9f500cb86bf4e62f1f8fc1dc17e89) shows that `reviewRegistration.html` has been essentially reverted to it's previous contents. All other changes are removals, and none of the removed methods are still used in the codebase, thus this PR is safe to merge.


hlbraddock:
In a subsequent PR, the backend will return a spouse registration from the `getCurrentRegistration()` (if no other registration available), depending on the particular conference configuration, as specified by the conference creator. For now, this simplifies the front end and the existing merge logic, which is timely given Cru 17 registration opens shortly. @OzzieOrca

PR #531, opened 2017-03-06
Disallow pages on HTTP or not on a subdomain

canac:
Ensure that in staging and production, the SPA is running on an SSL-secured page and that it is not running on the raw `eventregistrationtool.com` domain because our SSL certificate is not valid for that domain.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -14,6 +14,8 @@
     </script>
   <![endif]>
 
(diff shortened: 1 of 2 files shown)


canac:
Yes, it does ideally belong on the server. But the front-end server is just serving static files, so we can't currently have redirection logic live there.

PR #532, opened 2017-03-06
Correctly build clientSideRedirection.js

canac:
Make sure that `clientSideRedirection.js` gets built into the `eventApp.js` bundle. Fixes bugs introduced in #531.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -12,8 +12,7 @@
     <script type="text/javascript">
       window.location = '/browserUnsupported.html';
     </script>
-  <![endif]>
-  -->
+  <![endif]-->
 
   <!-- Adobe DTM -->
   <!-- build:remove:dist -->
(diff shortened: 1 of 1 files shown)


app/index.html
```diff
@@ -13,8 +13,7 @@
       window.location = '/browserUnsupported.html';
     </script>
   <![endif]>
-
-  <script src="scripts/clientSideRedirection.js"></script>
+  -->
```
OzzieOrca: I don't see what this is ending http://www.quirksmode.org/css/condcom.html

canac: It's supposed to close the tag on line 11. But should I change line 15 to `<![endif]-->`? I'm not familiar with IE conditional comments. The problem is that currently, everything between line 11 and the end of the Adobe DTM comment is being commented out.

OzzieOrca: Ya I think `<![endif]-->` is better.

canac: ✔️  Fixed

PR #533, opened 2017-03-12
IM-203 - Allow Admins To Switch Payment Gateway Type to Merchant ID

dougworner:
Allow Admins To Switch Payment Gateway Type to Merchant ID

Currently, if you create a conference, you only have the option to add a merchant id for Payment Options Credit Card payment. This is what we want.

However, if you already have a conference that takes Authorize.NET credentials, we'd like the conference admin to be able to switch to Merchant ID.

Add to the Payment Options Credit Card section a Payment Gateway Type menu with two options, 'Authorize.NET' and 'TSYS Merchant ID', but only if the current payment gateway type is 'Authorize.NET'. Otherwise, the conference must stay as Merchant ID.


test/spec/controllers/eventDetails.js
```diff
@@ -1,65 +0,0 @@
-'use strict';
```
canac: Why are you deleting this file? Are the tests stale?

app/views/eventDetails/paymentOptions.html
```diff
@@ -5,6 +5,15 @@ <h2 class="page-title border" translate>Payment Options</h2>
     <h3 class="panel-title" translate>Credit Card</h3>
   </div>
   <div class="panel-body">
+    <div class="form-group">
+      <div class="col-sm-12 stacked-spacing-col-sm" ng-if="conference.paymentGatewayType === 'AUTHORIZE_NET'">
+        <label translate>Select A Payment Gateway Type:</label>
+        <select class="form-control" ng-model="conference.paymentGatewayType">
+          <option value="TSYS" translate>TSYS Merchant ID</option>
```
canac: I think you mean TSYS here.

app/views/eventDetails/paymentOptions.html
```diff
@@ -35,6 +44,7 @@ <h3 class="panel-title" translate>Credit Card</h3>
         </div>
       </div>
     </div>
+
```
canac: Can you remove this stray line-break?

app/views/eventDetails/paymentOptions.html
```diff
@@ -5,6 +5,15 @@ <h2 class="page-title border" translate>Payment Options</h2>
     <h3 class="panel-title" translate>Credit Card</h3>
   </div>
   <div class="panel-body">
+    <div class="form-group">
+      <div class="col-sm-12 stacked-spacing-col-sm" ng-if="conference.paymentGatewayType === 'AUTHORIZE_NET'">
+        <label translate>Select A Payment Gateway Type:</label>
+        <select class="form-control" ng-model="conference.paymentGatewayType">
```
canac: These options could be generated by the data in `$scope.paymentGateways`. I won't say that you have to, but it would reduce duplication of the labels. You would use `ng-options` to specify an array of options generated from `$scope.paymentGateways`. There would be one option per entry, with the key being the option value and the `name` field being option label. Let me know if you decide to go that route and need help figuring it out.

app/views/eventDetails/paymentOptions.html
```diff
@@ -61,57 +71,58 @@ <h3 class="panel-title" translate>Checks</h3>
       </div>
       <div class="col-sm-3 stacked-spacing-col-sm">
         <select class="form-control" ng-model="conference.checkMailingState">
-          <option value="AL">Alabama</option>
```
canac: Great job catching these untranslated labels! 👍

test/spec/controllers/eventDetails.js
dougworner: I was having issues with Intellij and debugging. instead of using the actual eventDetails.js file when setting breakpoints it was the test/spec/controllers/eventDetails.js file.  So if I made a code change and then refreshed browser the breakpoints would not be accurate.  So I was forced to stop, save, restart every time I made a change which took longer. If there is a better approach please let me know.

app/views/eventDetails/paymentOptions.html
dougworner: Thanks.  I did some research to make sure we could use translate with and select option with gettext but really could not find a lot of documentation with this specific use case.  But also didn't find anything against.

app/views/eventDetails/paymentOptions.html
dougworner: I was going strictly off of what IM-203 Jira says, "two options, 'Authorize.NET' and 'TSYS Merchant ID'."
So I used 'Authorize.NET' and 'TSYS Merchange ID'.
Do you think it would be better for end users to make this select option 'TSYS'?

app/views/eventDetails/paymentOptions.html
dougworner: Think I got it.

app/views/eventDetails/paymentOptions.html
dougworner: I think that is a good suggestion (less code).  I made the change, but waiting to hear back from Lee as the link I used to use at the bottom of the https://stage.eventregistrationtool.com/#/ site is now missing (i.e. ‘Local Dev Environment’).
I would always click on this link after login to Cru Relay to do dev/test locally.
Waiting to push changes until this is resolved.

app/views/eventDetails/paymentOptions.html
dougworner: if going with the ng-options it will end up being 'TSYS' since will be using the name field.

hlbraddock:
@dougworner Can you incorporate changes as per @canac 's comments? Thanks!

test/spec/controllers/eventDetails.js
canac: It's fine if you keep the file deleted locally, just don't commit that deletion. Please revert this change. If you need help with the git commands, let me know.

app/views/eventDetails/paymentOptions.html
canac: From the other places it is used, it looks like putting the translate attribute on an element translates the text content of that element. Since it is the text content of the option elements that we want to translate, I think you are good to do it that way.

app/views/eventDetails/paymentOptions.html
canac: 👍

app/views/eventDetails/paymentOptions.html
canac: Yes, I do. A payment gateway of "TSYS Merchant ID" makes no sense to me. We can check with @hlbraddock, but I bet it was just a typo in the JIRA ticket.

dougworner:
@hlbraddock thanks for catching and "removing display of gateway type".  Test code I left in.

test/spec/controllers/eventDetails.js
dougworner: This happened before I completely recreated local environment with Git history of deleting this file.  So I recreated the file and pushed branch.

app/views/eventDetails/paymentOptions.html
dougworner: I agree.  We can double check with @hlbraddock if you think it is needed.

canac:
Here's my suggestion. Save a reference to `conference` in `$scope.originalConference`. Also, whenever the conference is successfully saved, update `$scope.originalConference`. Now `$scope.originalConference` will always refer to the conference data as stored on the server, and `$scope.conference` will refer to the conference with any unsaved changes. Then in `paymentOptions.html`, we can update the `ng-if` condition to only show the payment gateway type selector when `$scope.originalConference === 'AUTHORIZE_NET'`.

canac:
@dougworner Lee wanted the change that I suggested in my comment above implemented quickly, so I went ahead and made it.

hlbraddock:
Great work @dougworner !

dougworner:
Thank you @hlbraddock .
Sorry had to run to bank this morning otherwise I would have implemented Caleb's change.
So thankful to be able to help you out!

dougworner:
@canac Thanks for making the last change. Your suggestion makes sense and be one I remember for next time.

dougworner:
@canac Question to better understand...is the reason $scope.conference represents unsaved data because this object is not updated upon clicking the save button?  But if a person did a screen refresh it would be updated?

canac:
@dougworner `$scope.conference` reflects whatever data is displayed in the forms because it is bound to the form elements via `ng-model`. If the user changes the conference name, `$scope.conference.name` will contain the new name, but the old name will still be in the database because it has not been saved yet. `$scope.originalConference.name` will match the database value: the old conference name. If the user did a page refresh before saving, changes to `$scope.conference` will be lost. Does that answer your question?

dougworner:
@canac Yah, that makes it more clear.  Because $scope.conference only exists in current memory but not in the database (i.e. because save button has not been clicked to push these changes to the database), then any change you would make on the page would be immediately updated.  Only after clicking the save button would changes in $scope.conference be pushed to database.  So menu option will continue to display until save button is clicked and $scope.conference has been updated.  So not updated upon screen refresh but upon clicking saved button.

PR #535, opened 2017-03-23
Stop saving invalid payment gateways

canac:
When saving a conference without a payment gateway type set, only set the type to TSYS if a payment gateway id is set.

### DIFF
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -44,12 +44,16 @@ angular.module('confRegistrationWebApp')
     $scope.originalConference = conference;
     $scope.conference = angular.copy(conference);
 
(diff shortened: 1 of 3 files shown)


hlbraddock:
This looks good. A couple of questions:

1. Have you tested in stage? 
2. What is the behavior if a conference already has payment gateway type 'TSYS' set, but no payment id set (as will be the case for a number of existing conferences)? Will it leave it as 'TSYS' or will the conference be updated to no payment gateway type?

canac:
@hlbraddock 

1. No, I have not tested in stage, but I can do that today.
2. The payment gateway type is never cleared, so it will stay set to TSYS, even if no id is set.

hlbraddock:
If tested, are we ready to deploy? @canac

PR #537, opened 2017-04-27
Upgrade to Angular 1.6

OzzieOrca:
Main changes are:
- Convert `$http`'s `success` to `then` and `error` to `catch`. The function signatures also changed from `onSuccess(data, status, headers, config)` to `onSuccess(response)`. See [1.6 changelog](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#http-due-to)
- Upgraded packages to get JSHint and karma working locally and fixed some JSHint errors
- Specified `$locationProvider.hashPrefix('');` as the [default was changed to `#!`](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#location-due-to)
- Update `$cookies` usage. [You can no longer access cookies with properties.](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#breaking-changes-16)
- Remove `//` from logout path replace and redirect by removing trailing slash from `/logout/`

This could use some more testing but I've fixed all the errors I've seen so far.


app/scripts/controllers/accessEvent.js
```diff
@@ -18,7 +18,7 @@ angular.module('confRegistrationWebApp')
           reasonForRequest: reasonForRequest,
           permissionLevel: 'REQUESTED'
         }
-      }).success(function () {
+      }).then(function () {
         $modalInstance.close();
       }).error(function (data) {
```
canac: This needs changed to `catch`, right?

OzzieOrca: Ah. I thought I had searched for all those. Guess I need to do it again and double check. Good catch :)

OzzieOrca: I hope that's the only one I missed. I didn't see any others.

OzzieOrca:
Idk what I did. I thought I was updating the handlers I changed to `catch` as I went. I must have not done it for some search. Thanks for finding that.

OzzieOrca:
Thanks! I don't expect you to :) I'll do some more testing and get it on stage sometime.

hlbraddock:
@OzzieOrca approved +1

app/scripts/controllers/registration.js
```diff
@@ -167,7 +167,7 @@ angular.module('confRegistrationWebApp')
       }), true);
     };
 
-    var findAnswersToSave = function(){
```
adammeyer: Curious why you changed this to a named function.

OzzieOrca: Mmm I think probably because jshint was complaining. I don't remember exactly. This is defined after its usage. JS will hoist a function declaration but will only hoist the variable part of the function expression and will define the value of it later. (At least that's my understanding, correct me if I'm wrong) Functionally I don't think this matters here as the functions using this function aren't called immediately.

OzzieOrca: This might have also motivated this change https://github.com/CruGlobal/conf-registration-web/pull/537/files/d78487aa8a65289b882e2b964439615581bd5291#diff-4d5aa81bf4f18104bb6ea53a8b5d1f43R11 so I didn't have to move this function to the top of the file. Changing that `latedef` config allows function definitions to be defined after their usages but still disallows function expressions to be defined after usage.

PR #541, opened 2017-06-13
Use webpack and upgrade vendor packages

OzzieOrca:
- Add webpack tooling with support for ES6, SCSS, caching, eslint, and codecov
- Remove grunt, bower, jshint
- Import rollbar from npm
- Upgrade uib-bootstrap, lodash, angular-ui-tree
- Update travis to deploy to S3
- Use angular-wysiwyg from github and add preview button back with a float
- Migrate from LESS to SCSS
- Use $route.reload() to refresh page after permissions check instead of location redirects
- Add ccp.js under components since yarn doesn't support zip packages
- Use $locationProvider.html5Mode with hashPrefix('') to redirect old routes and remove #/ from hrefs
- Remove unused e2e test files

I've got this at [cru-ert-web-stage.s3-website-us-east-1.amazonaws.com](cru-ert-web-stage.s3-website-us-east-1.amazonaws.com). [clientSideRedirection.js](https://github.com/CruGlobal/conf-registration-web/blob/2c48dd193d9c87807b39533227bb2a47557836d6/app/scripts/clientSideRedirection.js#L24) and the auth redirect make this hard to use. I haven't setup the prod bucket or Cloudfront yet. Maybe Cloudfront can handle this redirection.

I haven't gone through the app and tested every feature yet. I'll do that once I get staging setup better.

I decided now's the time to upgrade the tooling since we are moving to S3. I probably should have upgraded the vendor packages separately but I just went for it with the move from bower to yarn/npm.

I think I'll save https://github.com/CruGlobal/conf-registration-web/pull/537 (Angular 1.6 upgrade) to go live with this PR so it gets a little more testing.

My current thought with deployment is to do development on `master` but don't deploy anything on a push to `master` and have a push to `staging` or `production` deploy to the corresponding environment. That way you can pick a commit to deploy for either environment by hard resetting that branch to a certain commit. Let me know what you think. That is super easy to switch around.


adammeyer:
Looks great @OzzieOrca!  You can remove the client redirect script, cloud front will do that redirection.  This will also make it usable on the s3 website domain.

I also think we'll want to add some long caching headers to the s3 files, minus index.html.

OzzieOrca:
https://d2py3bk8gadrkw.cloudfront.net/

.travis.yml
```diff
@@ -1,56 +1,40 @@
...
+- bash <(curl -s https://codecov.io/bash)
+cache:
+  yarn: true
+  directories:
+  - node_modules
+before_deploy:
+  - pip install --user awscli
+deploy:
+- provider: script
+  script: ~/.local/bin/aws s3 sync dist s3://cru-ert-web-stage --region us-east-1 --acl public-read --exclude '*.html' --cache-control 'public, max-age=31536000'
```
adammeyer: Looks great, should the `--delete` argument be added to keep things clean?  Also, I don't think the region is required, but it probably doesn't hurt.

OzzieOrca: I was debating that. If we delete the files but the user has an old HTML file, they could start getting 404s. Cloudfront may cache those long enough that we're safe. I guess there a risk that Cloudfront evicts something from the cache or 2 deploys happen quickly enough that something doesn't get cached. If we add delete, the prod cache control for the HTML files needs to be no cache which might be good to have anyways. Idk how Cloudfront regions work. If a user hits Cloudflare in a region that isn't warm does that request go to S3 or does Cloudflare share resources somehow?

Actually since I've got the S3 sync split into 2 requests, would a delete kill the files that are excluded? That might be a problem.

adammeyer: If the file isn't in cloudfront, it will hit s3 directly.  o yes, the separate requests would be a problem.  It would probably delete files from the previous request.  I guess it's fine.  Just might have many files over time.

OzzieOrca: It would be nice if we could delete everything that wasn't in the last 2 deploys but idk how to implement that. We could add `--delete` to the staging environment since that shouldn't matter much.

PR #544, opened 2017-06-21
Integrate Adobe Analytics with Angular

OzzieOrca:
- Refactor angular config and run into seperate files
- Move vendor imports to app.module.js
- Add analytics service
- Add digitalData layer
- Send 'page view' direct call rule on $routeChangeSuccess
- Add track calls to landing and reviewRegistration pages

I probably should have done my module/config refactoring in another PR but I was adding to the run config and decided it needed to be cleaned up.

This has an issue where the page titles sent to analytics are one page behind because `$routeChangeSuccess` fires before the controller is instantiated which is where the titles are declared. We probably need to do something like https://github.com/CruGlobal/mpdx_web/blob/staging/src/routes.js#L14. Idk what ngRoute supports. I just haven't gotten to that yet. Maybe that should happen in another PR.

@adammeyer Can you make sure I didn't miss anything from https://github.com/CruGlobal/conf-registration-web/pull/524? Thanks :) Sorry that branch got way behind with my upgrades.


OzzieOrca:
@brenton05 This is on https://stage.eventregistrationtool.com/ if you want to test it. As mentioned, the page titles are one page behind and that needs to be fixed in Angular. I also couldn't figure out how to make the debugger load the `loggedInStatus`. It is updating in `digitalData`.

I copied the `page view` direct rule call from MPDX. I made a few other changes to the DTM config.

app/scripts/controllers/landing.js
```diff
@@ -44,6 +45,11 @@ angular.module('confRegistrationWebApp')
         };
         $scope.searchComplete = val;
         $scope.searchResults = response.data;
+
+        analytics.digitalData.searchFilter = '';
```
OzzieOrca: @adammeyer @brenton05 What is supposed to be here? Is it necessary to send this with an empty string? Did this just not get finished in the other PR?

brenton05: @adammeyer @OzzieOrca this was used on the Give site to capture the name of the search filters if someone clicks on them. It tells us if people interact with the filters to help them find what they are looking for. For ERT it could be used to capture the names of the "locations" or "date" options when searching but right now we don't have to worry about that. Let's leave it in the data layer though.

brenton05: @OzzieOrca as for testing ERT I was able to begin testing. So far I don't see the page view direct call firing on the home page unless I'm logged in. Also, SSO Guid is not exposed in the data layer when I'm logged in. The logged In data element is coded to populate logged in or logged out based on SSO Guid being exposed. It wasn't working because when copied from MPDX the direct call had a lowercase "l" in it but it's fixed now. It will always say not logged in until we expose SSO. Or I could change the data element rule to look for logged in status rather than reply on SSO being populated. Maybe I'll change that now and test it. I'm free this afternoon to chat more if needed (6/21).

brenton05: @OzzieOrca I changed the Logged In Status data element to custom script and it's now pulling the status. It's no longer relying on SSO Guid to be populated. Will it say "Logged In" no matter which method a user logs in by (FB, relay, etc)?

`if (digitalData.user["0"].profile["0"].profileInfo.loggedInStatus == 'Logged Out') {
      return "Not Logged In";
  } else {
      return "Logged In";
  }`

OzzieOrca: Currently the backend isn't sharing the SSO Guid, GR ID, or Facebook ID so those are placeholders for now. I do have access to the login provider (`RELAY` or `FACEBOOK`). Where would you like me to put that in digitalData?

I'll fix the only tracking on logged issue and am looking into adding the other filters.

app/scripts/controllers/landing.js
```diff
@@ -44,10 +45,26 @@ angular.module('confRegistrationWebApp')
...
+    $scope.filterSearchResults = function(){
+      return _.filter($scope.searchResults, event => {
+        return dateFilter(event) && locationFilter(event);
+      });
+    };
+
+    $scope.filterUpdate = function(){
+      analytics.digitalData.searchTerm = $scope.searchVal;
+      analytics.digitalData.searchLocationFilter = $scope.eventFilters.locationName;
+      analytics.digitalData.searchDateFilter = $scope.eventFilters.date;
```
OzzieOrca: @brenton05 I split these out into 2 variables. Let me know if you want me to name them something else or try to serialize them back into one variable (and what format you would want that in).

brenton05: @OzzieOrca I verified that the digitalData layer updates with the search term, location filter, and data filter when used. Seems like the pageview direct call rule fires each time something happens on ERT like a search. I'm thinking for the search interactions I could just listen for a change in the search data element. I would need to test that. 

Also, I'm not getting SSO Guid or pagenames. All pages names in the data layer are the same. See attachment for details. I did however get the logged in status to work.

screen shot 2017-06-29 at 9 38 41 am

brenton05: @OzzieOrca you get a chance to look at this?

OzzieOrca: SSO GUID and GR ID are not provided by the API so I can't hand them to you at the moment. I could tell you if the user signed in with Facebook or Relay if that is useful but right now I don't have any IDs on the front end.

Page titles still need work. That will probably happen in another pull request. Are you ok with this going live without page titles or should we wait? The page titles are currently defined in the wrong place to be useful for analytics.

The search direct call rule should be named `search`and should fire anytime seach parameters are changed but should not fire a `page view` https://github.com/CruGlobal/conf-registration-web/pull/544/files#diff-78cedf0c6c09b9633fef47656a98872aR64. If you are seeing both I guess I need to do some debugging.

brenton05: @OzzieOrca I don't think we should push to prod until we get page titles. I'll check out the search direct rule call. I didn't know one was firing. I can probably just based the search stuff on that direct call rule then.

OzzieOrca:
@adammeyer Can you review my page titles commit? Thanks :)

PR #548, opened 2017-06-27
EVENT-501

dougworner:
Disable ability to use Authorize.Net as payment.  Only TSYS is now allowed.


app/scripts/controllers/eventDetails.js
```diff
@@ -32,14 +32,9 @@ angular.module('confRegistrationWebApp')
     };
     $scope.changeTab($scope.tabs[0]);
 
+    //AUTHORIZE_NET no longer used.
+    //TSYS is now the only payment allowed.
```
OzzieOrca: I don't think these comments will continue to be useful. Seems like it's commenting on code that won't exist in the future. I think you can just toss them. TSYS is the only payment gateway listed below now.

app/views/eventDetails/paymentOptions.html
```diff
@@ -6,17 +6,12 @@ <h3 class="panel-title" translate>Credit Card</h3>
   </div>
   <div class="panel-body">
     <div class="form-group">
-      <div class="col-sm-12 stacked-spacing-col-sm" ng-if="originalConference.paymentGatewayType === 'AUTHORIZE_NET'">
-        <label translate>Select A Payment Gateway Type:</label>
-        <select class="form-control" ng-options="key as value.name for (key, value) in paymentGateways" ng-model="conference.paymentGatewayType"></select>
+      <div class="col-sm-12 stacked-spacing-col-sm">
+        <label translate>Payment Gateway Type Used:  {{conference.paymentGatewayType !== null ? conference.paymentGatewayType : 'Other (not TSYS)'}}</label>
```
OzzieOrca: I'm wondering if it is useful to tell the user if their event is using TSYS or not. @hlbraddock do users interact with TSYS at all when requesting a `Merchant Account ID` or managing payments or is that intended to be a black box to the user?

app/scripts/controllers/paymentModal.js
```diff
@@ -326,6 +326,7 @@ angular.module('confRegistrationWebApp')
     /*
     Returns false is payment if payment and refund channel are credit card and current time is less than 24 hours from
     payment time.  If a refund is issued before the payment settles, authorize.net will do a full refund regardless of amount.
+    ***Authorize.net no longer used, but payment is still used so keeping code below for now.***
     */
     $scope.isPartialRefundAvailable = function(payment, refundChannel) {
```
OzzieOrca: @hlbraddock Do you know if the idea of partial refunds still exists for TSYS? This seems to have some specific rules preventing partial refunds for 24 hours for Authorize.net.

app/views/eventDetails/paymentOptions.html
hlbraddock: I doubt that they do. Likely they simply interact with their team's finance/operations leader, who works with Cru Finance department to get an assigned merchant account.

app/scripts/controllers/eventDetails.js
dougworner: I agree.  I removed them.

app/views/eventDetails/paymentOptions.html
OzzieOrca: Thanks Lee. @dougworner I think you can just remove this section. Sound reasonable?

dougworner: Assuming we should remove this then.  It has been removed.

app/scripts/controllers/paymentModal.js
OzzieOrca: @hlbraddock Did you see this? When does TSYS allow partial refunds, if at all?

app/views/eventDetails/paymentOptions.html
dougworner: I removed the section.

app/scripts/controllers/paymentModal.js
hlbraddock: Partial refunds exists with TSYS. I don't know that they have the same restrictions as Authorize.NET did. I think they don't.

OzzieOrca: So either make this function always return true or delete it and code where it is called? Do we need to test this somehow or just assume there are no restrictions?

hlbraddock: I think the restriction imposed by Authorize.NET, namely to not allow refunds (full or partial) within a certain period of time after the initial payment, does not apply at all to TSYS. I think that if the refund request is initiated within a certain window of time, say where the transaction has not yet settled, then the transaction will be voided. Otherwise, the transaction will be returned. 

Therefore, any ERT code which applies the no longer applicable business rule can be removed.

dougworner:
@OzzieOrca I removed code related to time period.

OzzieOrca:
@dougworner We want to always allow partial refunds so that function would need to always return true. I wasn't sure where that function was actually used so I jumped into the code to see if I could clean it up any more (since a function that always returns true seemed pointless) and added a commit. Can you review my commit? Does that make sense? Let me know if I missed something.

OzzieOrca:
Eh looks like there are some tests that are going to fail. I didn't realize that function was tested.

dougworner:
@OzzieOrca Yah, you removed the code completely.  Along with the tests.  The reason I didn't was because only part of the logic pertained to the time period where the other part did not.
if(payment.paymentType === 'CREDIT_CARD' && refundChannel === 'CREDIT_CARD')
But if the function always returns true then makes sense to remove.
I searched rest of project and didn't find anything else for isPartialRefundAvailable.
So should be good.

OzzieOrca:
Ya if we take
`if(payment.paymentType === 'CREDIT_CARD' && refundChannel === 'CREDIT_CARD' && diff < lengthToSettle) {`
and now say that `lengthToSettle` is `0` since it is immediately settled in the sense that is it is settled enough to allow partial refunds then we get
`if(payment.paymentType === 'CREDIT_CARD' && refundChannel === 'CREDIT_CARD' && diff < 0) {`
which is the same as
`if(payment.paymentType === 'CREDIT_CARD' && refundChannel === 'CREDIT_CARD' && false) {`
which is the same as
`if(false) {`
which will never be run so the whole function returns `true`;

PR #549, opened 2017-07-12
EVENT-455

dougworner:
Do not display pages with no questions when user is registering for an event.

### DIFF
--- a/app/scripts/controllers/registration.js
+++ b/app/scripts/controllers/registration.js
@@ -18,18 +18,30 @@ angular.module('confRegistrationWebApp')
     $scope.savingAnswers = false;
 
     $scope.activePageId = pageId || '';
-    $scope.page = _.find(conference.registrationPages, { 'id': pageId });
-    $scope.activePageIndex = _.findIndex($scope.conference.registrationPages, { id: pageId });
-    $scope.nextPage = function(){
-      var visiblePageArray = _.filter($scope.conference.registrationPages, function(page) {
+    $scope.page = _.find(conference.registrationPages, {'id': pageId});
+    $scope.activePageIndex = _.findIndex($scope.conference.registrationPages, {id: pageId});
+    $scope.nextPage = function () {
+      var visiblePageArray = _.filter($scope.conference.registrationPages, function (page) {
         return $scope.pageIsVisible(page);
       });
-      return visiblePageArray[_.findIndex(visiblePageArray, { 'id': pageId }) + 1];
+      return visiblePageArray[_.findIndex(visiblePageArray, {'id': pageId}) + 1];
     };
 
     //if current page doesn't exist, go to first page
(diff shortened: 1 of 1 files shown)


app/scripts/controllers/registration.js
```diff
@@ -18,20 +18,32 @@ angular.module('confRegistrationWebApp')
...
     //if current page doesn't exist, go to first page
-    if($scope.activePageIndex === -1 && angular.isDefined($scope.page)){
+    if ($scope.activePageIndex === -1 && angular.isDefined($scope.page)) {
       $location.path('/' + $rootScope.registerMode + '/' + conference.id + '/page/' + $scope.conference.registrationPages[0].id);
     }
 
+    //If page contains no questions
+    if (angular.isDefined($scope.page) && $scope.page.blocks.length === 0) {
+      // If not last page, then advanced to next page
+      if ($scope.page.position !== ($scope.conference.registrationPages.length - 1)) {
```
OzzieOrca: Doing `<` might be safer since it would handle an out of range value for `$scope.page.position` but hopefully that is unlikely.

dougworner: I agree. Don't want any out of bounds issues.

dougworner:
@OzzieOrca I pushed to staging and did a quick smoke test.  Look good.  Thanks for the feedback!

OzzieOrca:
Looks good. Feel free to deploy it to prod. I'm a little confused with what you did to the branch. Looks like you merged it into itself?

dougworner:
I did merge it into itself.  Using git in intellij is still a bit of a learning curve for me.  Never deployed to prod so not sure of the steps. I tried pushing to master but don't think I have permissions. And the Merge pull request button is greyed out for me.

OzzieOrca:
Usually if I modify a commit on a local branch, say `EVENT-455`, I force push it to `origin/EVENT-455`. Just don't force push to `origin/master` (which I don't think the permissions on Github will allow you to do anyways).

I think I gave you permission to merge PRs if the status checks pass and you get approval. Merging to master is all you have to do for it to deploy to production. After merging you can watch the build and deploy at https://travis-ci.org/CruGlobal/conf-registration-web.

dougworner:
I don't use force that much because it always scares me of overwriting something that shouldn't be.  When ever I make a change in a local branch (e.g. EVENT-455) I usually will just push to same remote branch (e.g. EVENT-455). You are correct. I can't push to remote origin/master because it is prevented.  I merged the pull request with prod and looks like Travis completed.

OzzieOrca:
Haha ya that's a consideration. Whatever you are comfortable with. It should be fine if you are force pushing to your own branches. I seem to like amending the most recent commit if it has a bug in it and doing rebases and those modify commits which then need to be force pushed.

Yep looks like it is deployed 👍

PR #550, opened 2017-07-31
Fix discard changes redirect on leaving event details page

OzzieOrca:
@dougworner Can you review this? You can reproduce this by going to the event details page, changing something, and then clicking a link to go to another page. It was redirecting to the home page instead of the desired page.

I saw this solution https://stackoverflow.com/a/31061068/665224

### DIFF
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -54,8 +54,9 @@ angular.module('confRegistrationWebApp')
       return $scope.conference.paymentGatewayType;
     };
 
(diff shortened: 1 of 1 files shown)


dougworner:
@OzzieOrca I pulled the code and tested locally.  Every time I made a change I did receive the popup asking to either discard changes or stay on page.  If I selected discard it brought me to the page I selected and not the home page.  Post by user to save path into a new variable seems like its working.

OzzieOrca:
Thanks :)

PR #551, opened 2017-07-31
EVENT-449 - Event Registration Tool Search Slow

dougworner:
-change from _.throttle to _.debounce
-change _.debounce wait/delay
-only keep valid responses (val === $scope.searchVal)

### DIFF
--- a/app/scripts/controllers/landing.js
+++ b/app/scripts/controllers/landing.js
@@ -28,7 +28,7 @@ angular.module('confRegistrationWebApp')
     ];
     $scope.jumboImg = jumboImgs[_.random(0, jumboImgs.length - 1)];
 
-    $scope.eventSearch = _.throttle(function(val) {
+    $scope.eventSearch = _.debounce(function(val) {
       if(!val){ return; }
       if(val.length < 2){
         return;
@@ -38,16 +38,18 @@ angular.module('confRegistrationWebApp')
           conferenceName: val
         }
       }).then(function(response){
-        $scope.eventFilters = {
-          locationName: '',
-          date: ''
-        };
-        $scope.searchComplete = val;
-        $scope.searchResults = response.data;
+        if(angular.isDefined($scope.searchVal) && val === $scope.searchVal) {
+          $scope.eventFilters = {
+            locationName: '',
+            date: ''
+          };
+          $scope.searchComplete = val;
+          $scope.searchResults = response.data;
 
-        $scope.filterUpdate();
+          $scope.filterUpdate();
+        }
       });
-    }, 500, {leading: false});
+    }, 1000);
 
     $scope.filterSearchResults = function(){
       return _.filter($scope.searchResults, event => {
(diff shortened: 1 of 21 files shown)


dougworner:
Not sure if I have the logic correct or in the right place for val===searchVal.
Also wasn't sure what time to increase the delay to so just chose 1000.
Tested locally fine.

app/scripts/controllers/landing.js
```diff
@@ -28,26 +28,28 @@ angular.module('confRegistrationWebApp')
...
+          $scope.searchComplete = val;
+          $scope.searchResults = response.data;
 
-        $scope.filterUpdate();
-      });
-    }, 500, {leading: false});
+          $scope.filterUpdate();
+        });
+      }
+    }, 1000, {leading: false});
```
OzzieOrca: I think `leading` is false by default: https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L10338

app/scripts/controllers/landing.js
```diff
@@ -28,26 +28,28 @@ angular.module('confRegistrationWebApp')
...
-          conferenceName: val
-        }
-      }).then(function(response){
-        $scope.eventFilters = {
-          locationName: '',
-          date: ''
-        };
-        $scope.searchComplete = val;
-        $scope.searchResults = response.data;
+      if(angular.isDefined($scope.searchVal) && val === $scope.searchVal) {
```
OzzieOrca: So at this point, I think essentially no time has passed and this will always be true. Where we run into problems is in the `then` handler below where a request has been in flight for a while and may finish out of order. It is there we want to check to make sure the response is answering the correct search query before showing that response to the user.

If you want to try and abort the previous request too you could look into using the `timeout` config https://docs.angularjs.org/api/ng/service/$http#usage.

app/scripts/controllers/landing.js
dougworner: I think you are right. Based on the GitHub link you posted I see leading is initially set to false.  Then later checked if options is an object and then setting to leading = !!options.leading.  I tested what would happen if I just left out the {leading: false} from the _.debounce function and seemed to operate just the same.  So I removed it.

app/scripts/controllers/landing.js
dougworner: I tested both ways and seemed to operate the same whether I placed around the HttpPromise or within '''then'''.  I placed it within '''then'''.

dougworner: I tested both ways and seemed to operate the same whether I placed around the HttpPromise or within 'then'.  I placed it within 'then'.

OzzieOrca:
It might be good to add a loading indicator as part of fixing this issue. Also 1s might be too long but we'll see how it performs on prod.

dougworner:
@OzzieOrca  It seems to be performing a little quicker.  Can you confirm?
Also do you have a recommended load indicator we could use?

OzzieOrca:
Ya looks good.

Mmm [here's one](https://github.com/CruGlobal/conf-registration-web/blob/5c3c387ddb4bdd2a878d5deb9fb799622aebfe4d/app/index.ejs#L60) that's already being used. You can see it when changing to the `/eventRegistrations` page. [Here are docs](http://fontawesome.io/examples/) on making it larger

PR #552, opened 2017-08-10
Remove CCP libraries

OzzieOrca:
- Remove ccp.js and jsencrypt

Leftover files from https://github.com/CruGlobal/conf-registration-web/pull/548


OzzieOrca:
Yes [`cru-payments`](https://github.com/CruPCI/cru-payments/blob/master/package.json) uses both of them.

It uses `jsencrypt` for both credit cards and bank accounts but only uses CCP for bank accounts.

Since the ERT [only](https://github.com/CruGlobal/conf-registration-web/blob/master/app/scripts/directives/payment.js#L1) [uses](https://github.com/CruGlobal/conf-registration-web/blob/master/app/scripts/services/payment.js#L1) credit cards it only imports the `cru-payments-cc` version which excludes bank acount code. The source files for the bank account stuff is still in node_modules. And jsencrypt is mentioned in yarn.lock since it is a dependency of `cru-payments`.

The bank account stuff is still sent to Cru's CCP (credit card processing) server for fun legacy reasons which makes the naming confusing.

Good questions :) I tried to abstract stuff away into cru-payments.

app/scripts/app.module.js
```diff
@@ -19,7 +19,6 @@ import 'pickadate/lib/picker.date.js';
 import 'pickadate/lib/compressed/themes/default.css';
 import 'pickadate/lib/compressed/themes/default.date.css';
 import 'angular-environment/dist/angular-environment.js';
-import 'cru-payments/dist/cru-payments-cc.js';
```
OzzieOrca: I also saw and removed this since it is specifically imported where needed.

dougworner:
Didn't find anything else.  Thumbs up.

PR #554, opened 2017-08-11
Enable subresource integrity and minify root html files (.ejs templates)

OzzieOrca:


OzzieOrca:
The basic premise of webpack is you give it an [entry point](https://github.com/CruGlobal/conf-registration-web/blob/5013565d30681ae2af341a3a34c6cbb0ebd2c420/webpack.config.js#L35-L37) that imports everything else your app needs and then you configure [a bunch of loaders](https://github.com/CruGlobal/conf-registration-web/blob/5013565d30681ae2af341a3a34c6cbb0ebd2c420/webpack.config.js#L105-L186) that tell it what to do when it tries to import each file, depending on the file type. Some loaders we're using are:

- `babel-loader` - compiles ES6 code to browser compatible JS code
- `html-loader` - minifies html and looks for assets included in the html and loads those also
- `image-webpack-loader` - compresses images and passes them on to the `file-loader`
- `file-loader` - emits the imported object as a file and to returns its public URL
- `sass-loader` - compiles SCSS to CSS
- `css-loader` - minifies CSS and emits it similarly to the file loader

There is obvious a lot of other stuff going on but it nice that you can tell webpack to use a loader for a certain file type and it knows what it should do to package it up. There are also plugins you can add that do useful things.

The plugins I was modifying/adding were:
- [`HtmlWebpackPlugin`](https://github.com/CruGlobal/conf-registration-web/blob/5013565d30681ae2af341a3a34c6cbb0ebd2c420/webpack.config.js#L56-L59) - takes your entry html file and adds your compiled/bundled js/css to it. I was configuring it to minify the HTML too.
- [`SriPlugin`](https://github.com/CruGlobal/conf-registration-web/blob/5013565d30681ae2af341a3a34c6cbb0ebd2c420/webpack.config.js#L97-L99) - automatically adds [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) attributes to all included assets. I was doing some work with SSL to make the app PCI compliant and, unrelatedly, discovered Webpack could easily do this and figured it would be a good to add.

You can run [`npm run build`](https://github.com/CruGlobal/conf-registration-web/blob/5013565d30681ae2af341a3a34c6cbb0ebd2c420/package.json#L73) and webpack will compile and bundle everything and spit it out to the `/dist` folder if you want to see what it builds.

PR #556, opened 2017-08-17
EVENT-508

dougworner:
Add event search loading indicator to landing page.

### DIFF
--- a/app/scripts/controllers/landing.js
+++ b/app/scripts/controllers/landing.js
@@ -29,6 +29,7 @@ angular.module('confRegistrationWebApp')
     $scope.jumboImg = jumboImgs[_.random(0, jumboImgs.length - 1)];
 
     $scope.eventSearch = _.debounce(function(val) {
+      $rootScope.loadingMsg = 'Search Events...';
       if(!val){ return; }
       if(val.length < 2){
         return;
@@ -39,13 +40,14 @@ angular.module('confRegistrationWebApp')
         }
       }).then(function(response){
         if(angular.isDefined($scope.searchVal) && val === $scope.searchVal) {
+          $rootScope.loadingMsg = '';
           $scope.eventFilters = {
             locationName: '',
             date: ''
           };
+          $rootScope.loadingMsg = '';
           $scope.searchComplete = val;
           $scope.searchResults = response.data;
-
           $scope.filterUpdate();
         }
       });
--- a/app/scripts/controllers/landing.js
+++ b/app/scripts/controllers/landing.js
@@ -29,11 +29,10 @@ angular.module('confRegistrationWebApp')
     $scope.jumboImg = jumboImgs[_.random(0, jumboImgs.length - 1)];
 
     $scope.eventSearch = _.debounce(function(val) {
-      $scope.loadingSearchResults = true;
       if (!val || val.length < 2) {
-        $scope.loadingSearchResults = false;
         return;
       }
+      $scope.loadingSearchResults = true;
       $http.get('conferences', {
         params: {
           conferenceName: val
--- a/app/scripts/controllers/landing.js
+++ b/app/scripts/controllers/landing.js
@@ -49,7 +49,7 @@ angular.module('confRegistrationWebApp')
           $scope.filterUpdate();
         }
       });
-    }, 1000);
+    }, 800);
 
     $scope.filterSearchResults = function(){
       return _.filter($scope.searchResults, event => {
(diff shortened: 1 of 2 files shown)


OzzieOrca:
I don't think we need to block user interaction during searching. At the end of the debounce, it prevents the user from changing their search until the server comes back with a response.

I was thinking something more like this:

screen shot 2017-08-17 at 12 52 10 pm


```html
<div class="text-center">
  <h2>Searching events…</h2>
  <i class="fa fa-spinner fa-spin fa-4x"></i>
</div>
```

Idk, what do you think? You could try some other styles or throw it in a panel/well if that looks any better. Idk

dougworner:
I tested your code.  How would we control it so it doesn't spin continuously? As soon as I bring up the site the indicator is spinning.

I did some other testing as well you might find interesting.  I replaced $rootScope with $scope and it appears now that the user is able to interact.  I added more places to clear out the value since I noticed when I type in a value and then backspaced it was not being cleared.

What do you think?

app/scripts/controllers/landing.js
```diff
@@ -29,8 +29,12 @@ angular.module('confRegistrationWebApp')
     $scope.jumboImg = jumboImgs[_.random(0, jumboImgs.length - 1)];
 
     $scope.eventSearch = _.debounce(function(val) {
-      if(!val){ return; }
+      $scope.loadingMsg = 'Searching Events...';
+      if(!val){
+        $scope.loadingMsg = '';
+        return; }
       if(val.length < 2){
```
OzzieOrca: Can you consolidate this with the if above? Seems like they do the same thing.

OzzieOrca:
Umm you can toss the `fa-spin` class with `ng-class` if you don't want it to spin some of the time. Idk what you have in mind.

I guess I usually try to not add overlays unless they are really necessary. My idea would be to show the current initial content (Register for an event To register for an event, search...) when the user hasn't typed anything or has deleted their search, show the loading indicator in place of it during a search and then show the results when they come back.

dougworner:
Let me know if this is closer to what you were thinking.  I left fa-spin in for now.

dougworner:
Made one small change.

app/views/landing.html
```diff
@@ -32,6 +32,10 @@ <h1 translate>Search for an event</h1>
     </div>
   </div>
 </div>
+<div class="text-center" ng-if="loadingMsg">
+  <h2>{{loadingMsg}}</h2>
+  <i class="fa fa-spinner fa-spin fa-4x"></i>
+</div>
 <div class="container spacing-above-md" ng-if="!searchComplete">
```
OzzieOrca: I think we should hide this when showing the loading indicator.
Could do: `ng-if="!loadingMsg && !searchComplete"`

app/views/landing.html
```diff
@@ -32,6 +32,10 @@ <h1 translate>Search for an event</h1>
     </div>
   </div>
 </div>
+<div class="text-center" ng-if="loadingMsg">
+  <h2>{{loadingMsg}}</h2>
```
OzzieOrca: I think you can just hardcode `Searching Events&hellip;` here since the message never changes.

You can make it translatable by doing `<h2 translate>Searching Events&hellip;</h2>`.

Then `loadingMsg` could just be a flag (true/false). Can you think of a more descriptive variable name for it that implies it is for searching? Maybe `loadingSearchResults` or something.

app/scripts/controllers/landing.js
```diff
@@ -29,23 +29,24 @@ angular.module('confRegistrationWebApp')
...
+        $scope.loadingMsg = '';
         return;
       }
       $http.get('conferences', {
         params: {
           conferenceName: val
         }
       }).then(function(response){
+        $scope.loadingMsg = '';
         if(angular.isDefined($scope.searchVal) && val === $scope.searchVal) {
```
OzzieOrca: Disabling the loading indicator should be done in here cuz that's where we know we have results for the current search query and that `$scope.searchResults` has been set.

app/scripts/controllers/landing.js
```diff
@@ -29,8 +29,9 @@ angular.module('confRegistrationWebApp')
     $scope.jumboImg = jumboImgs[_.random(0, jumboImgs.length - 1)];
 
     $scope.eventSearch = _.debounce(function(val) {
-      if(!val){ return; }
-      if(val.length < 2){
+      $scope.loadingSearchResults = true;
```
OzzieOrca: If you move this to after the if statement that checks the length, you shouldn't have to deal with it in the if.

PR #558, opened 2017-09-06
Add ability for a conference to specify a custom CSS file for their r…

OzzieOrca:
…egistration pages

~~This doesn't have API support yet so the key name in the conference object may need to be changed.~~ Updated to match https://github.com/CruGlobal/ert-api/pull/721

FamilyLife wants to iFrame the ERT form on their conference info page and needs it to be custom branded. They are working on writing CSS overrides.

### DIFF
--- a/app/index.ejs
+++ b/app/index.ejs
@@ -36,6 +36,7 @@
 </head>
 
 <body ng-class="globalPage.bodyClass" ng-cloak>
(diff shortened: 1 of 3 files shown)


hlbraddock:
Ping me again sometime next week if I haven't provided the API yet for this. @OzzieOrca

PR #560, opened 2017-11-18
Update additional cost language

dsgoers:
Due to recent changes with FSG, the example of "t-shirt" is no longer acceptable.


dsgoers:
thx

PR #561, opened 2017-11-30
EVENT-524

dougworner:
Validate answer value for the Multiple Choice Question to prevent same answer multiple times.

### DIFF
--- a/app/scripts/directives/blockEditor.js
+++ b/app/scripts/directives/blockEditor.js
@@ -13,6 +13,7 @@ angular.module('confRegistrationWebApp')
         $scope.showClearBtn = true;
         $scope.isAdmin = true;
         $scope.ruleTypeConstants = ruleTypeConstants;
+        $scope.editBlockAddOptionAnswer = { value: '' };
 
         $scope.popup = {
           titleTemplateUrl: popupHyperlinkInformationTemplate
@@ -116,12 +117,18 @@ angular.module('confRegistrationWebApp')
           if (angular.isUndefined($scope.block.content)) {
             $scope.block.content = { 'choices': [] };
           } else if(angular.isUndefined($scope.block.content.choices)){
-            $scope.block.content.choices = [];
+              $scope.block.content.choices = [];
           }
-          $scope.block.content.choices.push({
-            value: newOption,
-            desc: ''
-          });
+          if (_.some($scope.block.content.choices, ['value', newOption])){
+            modalMessage.error('Error: You may not use the same answer more than once for this question. Each answer needs to be unique.');
+          }
+          else {
+            $scope.block.content.choices.push({
+              value: newOption,
+              desc: ''
+            });
+            $scope.editBlockAddOptionAnswer.value = '';
+          }
         };
 
         $scope.editBlockOptionMoveUp = function (index) {
--- a/app/views/components/blockEditor.html
+++ b/app/views/components/blockEditor.html
@@ -75,9 +75,9 @@
           </div>
 
           <div class="input-group spacing-above-md">
-            <input ng-model="editBlockAddOptionValue" type="text" placeholder="New Answer" class="form-control" ng-enter="editBlockAddOption(editBlockAddOptionValue); editBlockAddOptionValue='';">
+            <input ng-model="editBlockAddOptionAnswer.value" type="text" placeholder="New Answer" class="form-control" ng-enter="editBlockAddOption(editBlockAddOptionAnswer.value)">
                   <span class="input-group-btn">
-                    <button type="button" class="btn btn-success" ng-click="editBlockAddOption(editBlockAddOptionValue); editBlockAddOptionValue='';"><span class="glyphicon glyphicon-plus" page=""></span> <translate>Add</translate></button>
+                    <button type="button" class="btn btn-success" ng-click="editBlockAddOption(editBlockAddOptionAnswer.value)"><span class="glyphicon glyphicon-plus" page=""></span> <translate>Add</translate></button>
                   </span>
           </div>
 


app/scripts/directives/blockEditor.js
```diff
@@ -116,12 +116,17 @@ angular.module('confRegistrationWebApp')
           if (angular.isUndefined($scope.block.content)) {
             $scope.block.content = { 'choices': [] };
           } else if(angular.isUndefined($scope.block.content.choices)){
-            $scope.block.content.choices = [];
+              $scope.block.content.choices = [];
+          }
+          if (_.some($scope.block.content.choices, ['value', newOption])){
+            modalMessage.error('Error: Cannot use the same Multiple Choice Answer more than once.');
```
OzzieOrca: It looks like this code applies to multiple choice, checkbox, and dropdown questions. Can you make the message more generic? Maybe `You may not use the same answer more than once for this question. Each answer needs to be unique.`. Idk. I think I tend to make things more verbose haha.

Also `Error:` seems duplicate:
screen shot 2017-11-29 at 7 55 49 pm

If you end up doing inline validation when editing existing fields, it might make sense to do inline validation here too.

dougworner:
I'm not sure how to address this either.  Looking at the documentation for ng-messages you need to have a <form> tag with 'name' attribute but I could not find <form> declared.
https://docs.angularjs.org/api/ngMessages/directive/ngMessages

I guess we could use ng-if but not sure that would be the best.

thoughts?

dougworner:
Not sure how to address this either.  Looking at documentation for ng-messages you need to have a 'form' tag that also has a 'name' attribute filled in but I don't see this being used on blockEditor.html.  Not sure if using ng-if or ng-show would be the way to go.  thoughts?

https://docs.angularjs.org/api/ngMessages/directive/ngMessages

OzzieOrca:
Hmm. I'm thinking we just deploy the fix for validating the add new answer field and do the edit validation in another PR/issue. Can you fix my 2 comments? Did those make sense?

app/scripts/directives/blockEditor.js
dougworner: I modified the error to use your more descriptive message.

app/scripts/directives/blockEditor.js
```diff
@@ -116,12 +116,17 @@ angular.module('confRegistrationWebApp')
...
+          }
+          else {
+            $scope.block.content.choices.push({
```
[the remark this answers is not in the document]
dougworner: I removed it and placed it within the function but didn't notice any affect on behavior when tested.  Then I removed it completely and still didn't notice any affect.  Let me know if you want me to add it back into the function otherwise we can leave it out.

dougworner:
Changes made based on your last 2 comments.

dougworner:
@OzzieOrca changes made based on your last 2 comments.

app/scripts/directives/blockEditor.js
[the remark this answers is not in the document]
OzzieOrca: Oh there's 2 of them. The other is here: https://github.com/CruGlobal/conf-registration-web/pull/561/files#diff-ed772b44f382f1184004ff83ef7994c2R80.

One gets called when you click the button (`ng-click`), the other gets called when you press enter (`ng-enter`).

Removing them both and moving them into the function should work. The input should be cleared only when the answer is successfully added.

app/scripts/directives/blockEditor.js
OzzieOrca: 👍

dougworner:
Good catch!  ignore paymentsModal.html commit. no changes were made.  I just didn't notice that IntelliJ had this file checked.

app/scripts/directives/blockEditor.js
```diff
@@ -116,12 +116,17 @@ angular.module('confRegistrationWebApp')
...
+              $scope.block.content.choices = [];
+          }
+          if (_.some($scope.block.content.choices, ['value', newOption])){
+            modalMessage.error('Error: You may not use the same answer more than once for this question. Each answer needs to be unique.');
+          }
+          else {
+            $scope.block.content.choices.push({
+              value: newOption,
+              desc: ''
+            });
```
OzzieOrca: `$scope.editBlockAddOptionValue='';` needs to be added back after this line so when you successfully create a new answer, it clears the new answer input.

app/views/modals/paymentsModal.html
```diff
@@ -436,6 +436,7 @@ <h4 translate>Refund Payment</h4>
           <select ng-model="newTransaction.paymentType" class="form-control" ng-change="setTransactionAmount()">
             <option value="" disabled selected translate>Choose one...</option>
             <optgroup label="Payment">
+              <option value="CREDIT_CARD" ng-if="conference.paymentGatewayKeySaved" translate>Credit Card</option>
```
OzzieOrca: This is probably from a merge commit. This needs to be gone. Lee removed it a few days ago.

OzzieOrca:
Also all your commits have the same commit message from some previous issue... I guess you don't need to change the past but it the future unique messages would make it easier to follow.

dougworner:
I have been amending the changes which is why the comments stay the same.  Going forward I won't amend but add a new comment each time.

This is strange when I add the $scope.editBlockAddOptionValue=''; to the function it doesn't work.
I tried placing it in several places within the function with same results.
But when I place it back into blockEditor.html it does work.
Not sure why.
Which explains now why someone place this where it originally was.

thoughts?

OzzieOrca:
It's been awhile since I've worked with $scope issues and directives/components that are not using controllerAs. It had me stumped for a while.

https://stackoverflow.com/a/16973319/665224 is the answer. ng-if creates a new scope and updates to a child scope don't update their parent's scope. Using and object instead works because it gets passed by reference.

I added a commit. Can you look over it? Thanks :)

dougworner:
Tested fine.

Have scope issues improved or been removed in Angular 2.x and higher?

OzzieOrca:
Ya using directives with `controllerAs` makes it a little better. Isolate scope makes it even better. Components which were introduced in Angular 1.5 make both of those easy to use. I think this code was written before any of that.

I don't think Angular 2 has a concept of scope. They've moved away from that terminology.

Looks like that line adding the credit card option was still there. I couldn't figure out what happened to the git history. It looked different on github and locally. I just squashed all the commits and rebased it onto master since the messages didn't mean a whole lot anyways.

PR #564, opened 2017-12-08
EVENT-531

dougworner:
Validate option answer value for uniqueness when editing answer values

### DIFF
--- a/app/scripts/directives/blockEditor.js
+++ b/app/scripts/directives/blockEditor.js
@@ -23,6 +23,15 @@ angular.module('confRegistrationWebApp')
           $scope.answer = {};
         }
 
+        $scope.validateAddOptionAnswer = function (newOption) {
+          if (_.some($scope.block.content.choices, ['value', newOption])){
+            $scope.questionValidationMsg = 'Error: You may not use the same answer more than once for this question. Each answer needs to be unique.';
+          }
+          else {
+            $scope.questionValidationMsg = '';
+          }
+        };
+
         // Migrate old paragraph content objects
         if ($scope.block.type === 'paragraphContent' &&
           angular.isDefined($scope.block.content) &&
--- a/app/views/components/blockEditor.html
+++ b/app/views/components/blockEditor.html
@@ -75,7 +75,7 @@
           </div>
 
           <div class="input-group spacing-above-md">
-            <input ng-model="editBlockAddOptionAnswer.value" type="text" placeholder="New Answer" class="form-control" ng-enter="editBlockAddOption(editBlockAddOptionAnswer.value)">
+            <input name="editBlockAddOptionAnswer.value" ng-model="editBlockAddOptionAnswer.value" type="text" placeholder="New Answer" class="form-control" ng-enter="editBlockAddOption(editBlockAddOptionAnswer.value)" ng-change="validateAddOptionAnswer(editBlockAddOptionAnswer.value)">
                   <span class="input-group-btn">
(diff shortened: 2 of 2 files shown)


dougworner:
@OzzieOrca I looked at this for quite some time this evening. I tried using ng-messages and also creating a custom directive and some other examples on the web but could get none of them working.  The way scope is used in ERT gets confusing and hard to follow.  And when ng-controller and other standard methods of dealing with scope is not used adds more confusion.  This is the best solution I could come up with.

OzzieOrca:
You're code is a good simplistic way to do this. Unfortunately it doesn't look like it actually prevents the user from saving an invalid value, it just warns them.

You could try something like this https://www.webpackbin.com/bins/-L-qiNHomA-NBrC6tFMb. I wrote it with a little bit newer code (using a component with controllerAs: `$ctrl`). If you want to use it in the existing code, you need to drop the `$ctrl` everywhere.

The showErrors.js directive is already in the ERT repo. I've done it differently in a new repo by using a filter (https://github.com/CruGlobal/give-web/blob/master/src/common/filters/showErrors.filter.js) to check for `$invalid` and `$touched`.

dougworner:
Thx.  Even if I drop the $ctrl your code still uses the form name.  What is the form name for blockEditor.html? I'm not seeing it.  Looking for a <form> tag.  Looking in Chrome F12 developer tools I see the following but no name attribute:
<form ui-tree-nodes="" role="form" ng-model="conference.registrationPages" drop-type="page" novalidate="" class="ng-pristine ng-untouched ng-scope angular-ui-tree-nodes ng-not-empty ng-invalid ng-invalid-required ng-valid-email">

OzzieOrca:
I think you'll have to add it

dougworner:
My attempt at implementing ngMessage.  In my testing I was unable to save an invalid value (i.e. duplicates). Also, when I didn't find adding show-errors added anything noticeable.  This was a bit of a learning experience for me (ngMessage).  Spent several hours again this evening working on this.

dougworner:
@OzzieOrca Not sure if you saw my last post for review.

OzzieOrca:
I thought I had replied. Sorry about that.

`show-errors` should color the input red when it is invalid. But you need to add an angular validator for it to change the $error state.

The input also needs to have the `.form-control` class and be surrounded by something with the `.form-group` class for `show-errors` to change the styling.

dougworner:
@OzzieOrca I'm stuck on adding the angular validator.  What am I missing?  Just posted most recent changes.

dougworner:
@OzzieOrca did you see my last comment?

OzzieOrca:
Ya I saw it and spent a little time on it. I couldn't seem to get a hold of the ngModelController I wanted. I've been busy jumping into a new project so haven't gotten back to it.

dougworner:
@OzzieOrca I think that must have been my issue as well but just didn't realize it.  Could not get it to hook up to a validator.  I'll take another look at it.  We can always go back to what I had originally which was working but using a different approach.

dougworner:
@OzzieOrca I wasn't able to get it to work either so I approached this differently.  I'm still using ngMessages but I'm using ngModel $error object with ngMessages for validation.  When I tested it seems to work the way we want (i.e. prevents entering a duplicate value and displays a message to the user).

dougworner:
@OzzieOrca just checking if you have time to review this or if you want to wait until next year.

hlbraddock:
Closing until task is prioritized.

PR #565, opened 2017-12-22
HS-185610 - Extend CC Expiration Years

wrandall22:
This PR adds expiration years to make 20 instead of 9, this matches what Amazon does.  Based on [this HS ticket](https://secure.helpscout.net/conversation/491743860/185610?folderId=320670).


wrandall22:
Note: I'm not sure if this feature is desired or not, just is the fix for the Helpscout issue.

wrandall22:
@ryancarlson any objections on this? Not sure if you still have a stake in ERT or not.

PR #566, opened 2018-01-11
EVENT-533 - Create birthdate question

wrandall22:
This PR creates the front-end parts of a predefined birthdate question. Perhaps we don't actually need a separate template for it, not sure, maybe we can re-use the dateQuestion template.

[Back-end PR](https://github.com/CruGlobal/ert-api/pull/730) needs to be pushed first.


app/scripts/directives/blocks.js
```diff
@@ -153,3 +154,11 @@ angular.module('confRegistrationWebApp')
         restrict: 'E'
       };
     });
+
+angular.module('confRegistrationWebApp')
+  .directive('birthDateQuestion', function () {
+    return {
+      templateUrl: birthDateQuestionTemplate,
```
OzzieOrca: Ya since these templates look identical, I would just use the `dateQuestionTemplate` here. If it has some additional functionality we can create a new version later.

app/views/components/blockEditor.html
```diff
@@ -35,6 +35,7 @@
     <gender-question block="block" ng-switch-when="genderQuestion"></gender-question>
     <date-question block="block" ng-switch-when="dateQuestion"></date-question>
     <year-in-school-question block="block" ng-switch-when="yearInSchoolQuestion"></year-in-school-question>
+    <birth-date-question block="block" ng-switch-when="birthDateQuestion"></birth-date-question>
```
OzzieOrca: I might even just reuse  the `date-question` component here. It can still have a type of `birthDateQuestion` but just renders a date question since there's no logic in either component (directive) and the templates are the same.

app/views/components/blockRegistration.html
```diff
@@ -16,6 +16,7 @@
     <number-question block="block" ng-switch-when="numberQuestion"></number-question>
     <gender-question block="block" ng-switch-when="genderQuestion"></gender-question>
     <date-question block="block" ng-switch-when="dateQuestion"></date-question>
+    <birth-date-question block="block" ng-switch-when="birthDateQuestion"></birth-date-question>
```
OzzieOrca: and here

wrandall22:
Yeah, months was a bit difficult to navigate, but years not so much (at least back to 1902). I went ahead and did the month selector, which made it easier.

PR #569, opened 2018-01-17
EVENT-534 - Create predefined dormitory question

wrandall22:
This creates a predefined dormitory question and relies upon back-end changes in https://github.com/CruGlobal/ert-api/pull/732

### DIFF
--- a/app/scripts/directives/blockEditor.js
+++ b/app/scripts/directives/blockEditor.js
@@ -56,7 +56,7 @@ angular.module('confRegistrationWebApp')
         }
 
         //mapping default value to answer model for showing in front end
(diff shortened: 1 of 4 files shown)


wrandall22:
It is to record the dormitory piece of information in the `people` table. Apparently, someone wanted it, or it wouldn't be in the table.  I would guess it would be for reconnecting students to their dorm's Bible studies when they got back to campus.

wrandall22:
My guess is that off campus students would just put "Off Campus" or something. Also, event coordinators can give a custom question title I believe.

OzzieOrca:
Got it. Sounds fine. The case where the user has no dorm seems a little vague but idk what would be better. Maybe the conference admin would set up a question before this and add a rule to conditionally show it.

OzzieOrca:
Just saw your last comment. The admin could put that in the description or something. Cool.

PR #570, opened 2018-01-23
EVENT-532 - Default to Profile Question Checked

wrandall22:
For profile questions, they should default to checked for being saved to the profile. I dug in and found that the `defaultProfile` value was not being passed in when dropping a new question onto the form, and thus, every question defaulted to not being a profile question.


wrandall22:
thanks!

PR #575, opened 2018-05-01
EVENT-544: Add cookie domain to share session cookies between www and…

OzzieOrca:
… apex domains

- Upgrade node-sass

https://jira.cru.org/browse/EVENT-544

The node-sass upgrade isn't necessary but I had to recompile it locally for node 10 and it grabbed a new version so I left it.

I was considering the redirect to www but it looks like to do it with AWS, you'd have to create an S3 bucket for that apex domain for the entire purpose of the redirect and get Cloudfront to apply HTTPS to it somehow. Seemed too convoluted.

This just sets the domain of all cookies set by the client to the apex domain so they can be shared between the apex domain and the www subdomain.

I think that regex test line was functioning properly and the sourcemap was weird. It compiles to `/^\/auth\/.*/.test(a.url()) || (t.put("intendedRoute", a.path()), ...`.

### DIFF
--- a/app/scripts/app.config.js
+++ b/app/scripts/app.config.js
@@ -12,7 +12,7 @@ import helpTemplate from 'views/help.html';
 import privacyTemplate from 'views/privacy.html';
 
 angular.module('confRegistrationWebApp')
-  .config(function ($locationProvider, $httpProvider, $qProvider, $routeProvider, envServiceProvider, $compileProvider, gettext) {
+  .config(function ($locationProvider, $httpProvider, $qProvider, $routeProvider, envServiceProvider, $compileProvider, $cookiesProvider, gettext) {
     $locationProvider.html5Mode(true).hashPrefix('');
     $httpProvider.useApplyAsync(true);
     $qProvider.errorOnUnhandledRejections(false);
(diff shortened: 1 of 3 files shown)


app/scripts/app.config.js
```diff
@@ -32,22 +32,27 @@ angular.module('confRegistrationWebApp')
...
           apiUrl: 'https://api.stage.eventregistrationtool.com/eventhub-api/rest/',
-          tsysEnvironment: 'production'
+          tsysEnvironment: 'production',
+          cookieDomain: 'stage.eventregistrationtool.com'
         },
         production: {
           apiUrl: 'https://api.eventregistrationtool.com/eventhub-api/rest/',
-          tsysEnvironment: 'production'
+          tsysEnvironment: 'production',
+          cookieDomain: 'eventregistrationtool.com'
```
hlbraddock: It's probably not important, but should this be:

`cookieDomain: 'eventregistrationtool.com'`

OzzieOrca: My understanding is subdomains can read a parent domain's cookies but not the other way around.

I found some article that referred me to  https://scripts.cmbuckley.co.uk/ which I played with a bit. Seems like that's how it functions.

hlbraddock: Have you tested this to see that it resolves the issue https://jira.cru.org/browse/EVENT-544?

hlbraddock:
Have you tested this to see that it resolves the issue https://jira.cru.org/browse/EVENT-544?

hlbraddock:
If you're confident it won't break anything, we can deploy to production and test there, and revert if necessary.

PR #578, opened 2018-05-02
MHP-544: Perform clientside redirection from apex production domain t…

OzzieOrca:
…o www

We had this code before. I didn't think it was needed since Cloudflare handled the https redirect. I wish Cloudfront/S3 could handle this without creating more buckets. It would be nice to support the apex domain and www but the `intendedRoute` cookie isn't shared and auth redirects are broken when moving between domains.

### DIFF
--- a/app/scripts/clientSideRedirection.js
+++ b/app/scripts/clientSideRedirection.js
@@ -1,0 +1,22 @@
+// Do not redirect from localhost
+if (window.location.hostname !== 'localhost') {
+
+  /*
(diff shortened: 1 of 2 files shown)


hlbraddock:
Was this code in production before? Have you tested somewhere?

OzzieOrca:
Yes it was. I stole it from [here](https://github.com/CruGlobal/conf-registration-web/pull/541/files#diff-a23f310732a5bb12ce55e68685d2df99) and removed the IIFE since it is inside an ES6 module now. It's running on staging fine.

PR #588, opened 2018-09-12
EVENT-555 Date dependent questions

emiliano1:


app/scripts/services/DateRangeService.js
```diff
@@ -0,0 +1,39 @@
...
+    this.emitChange = function (emitter, value) {
+      cache[emitter] = value;
+
+      angular.forEach(subscribed[emitter], function(callback) {
+        callback(value);
+      });
+    };
+
+    this.calculateDateRange = function (startDate, endDate) {
+      if (angular.isUndefined(startDate) || startDate === null || startDate === '') {
```
emiliano1: I  believe that this `angular.isUndefined(startDate) || startDate === null || startDate === ''` can be changed to be `!startDate`.

app/scripts/services/DateRangeService.js
```diff
@@ -0,0 +1,39 @@
...
+        return 1;
+      }
+      if (angular.isUndefined(endDate) || endDate === null || endDate === '') {
```
emiliano1: I believe that this `angular.isUndefined(endDate) || endDate === null || endDate === ''` can be changed to be `!endDate`.

PR #589, opened 2018-09-18
EVENT-573 No Logout Button when Facebook Login

emiliano1:


hlbraddock:
@emiliano1 Looks like the Travis build failed with `
The command "yarn run test -- --single-run" exited with 1.`

PR #591, opened 2018-09-19
EVENT-499 Fix Questions with forced choices

emiliano1:


emiliano1:
@hlbraddock on it since the weekend.

hlbraddock:
One other possible implementation for this issue would be to remove answers from the backend api where they were no longer relevant (i.e. when rules dependent answers were no longer valid). 

I don't know if you considered this possible implementation and/or if you think it worthwhile. Likely, if you've already resolved by way of the front end, a backend solution isn't necessary then. 

Just wanted to at least ask you about the possibility.

@emiliano1

emiliano1:
Yes, I thought about fixing this in the backend, but the frontend solution was easier.

PR #594, opened 2018-10-01
EVENT-468 Group registrations

emiliano1:


app/scripts/controllers/eventRegistrations.js
```diff
@@ -317,38 +317,17 @@ angular.module('confRegistrationWebApp')
 
       var showGroupDialogOptions = {
         templateUrl: showGroupModalTemplate,
+        controller: 'showGroupModalCtrl',
```
OzzieOrca: I still think it would be great if these didn't have to share $scope and could use an Angular 1.5 component which ties the controller and template together in one place.

app/scripts/controllers/showGroupModal.js
```diff
@@ -0,0 +1,30 @@
...
+    $scope.registrationTypeFull = function(type){
+      if(!type.useLimit){
+        return false;
+      }
+      if(!type.availableSlots){
+        return true;
+      }
+
+      // subtract currentRegistration's registrants from availableSlots. if there's no slot left, return true
+      if(type.availableSlots - _.filter($scope.currentRegistration.registrants, { 'registrantTypeId': type.id }).length <= 0){
```
OzzieOrca: I still think this line could be cleaned up. It does extra math just to compare it against 0 when it could be compared directly and the return true/false are duplicate. You could just return the result of the comparison which is a bool.

app/scripts/directives/showGroup.js
```diff
@@ -0,0 +1,55 @@
...
+
+      $ctrl.$onInit = function() {
+        $ctrl.groupName = $ctrl.resolve.groupName;
+        $ctrl.registrationId = $ctrl.resolve.registrationId;
+        $ctrl.conference = $ctrl.resolve.conference;
+        $ctrl.getRegistration = $ctrl.resolve.getRegistration;
+        $ctrl.getRegistrantType = $ctrl.resolve.getRegistrantType;
+        $ctrl.editRegistrant = $ctrl.resolve.editRegistrant;
+        $ctrl.deleteRegistrant = $ctrl.resolve.deleteRegistrant;
+        $ctrl.registerUser = $ctrl.resolve.registerUser;
```
OzzieOrca: Having to do all this is unfortunate. It's nice that this makes it explicit what variables you are pulling out of resolve though.

You could do
```js
const {
  groupName,
  registrationId,
  conference,
  getRegistration,
  getRegistrantType,
  editRegistrant,
  deleteRegistrant,
  registerUser
} = $ctrl.resolve;
```
to get them all as local variables but that doesn't help copy them onto $ctrl if you need them in the template.

I suppose you could clone everything from resolve onto $ctrl but that makes it less explicit and could lead to bugs if you don't know what's being copied.

app/scripts/directives/showGroup.js
```diff
@@ -0,0 +1,55 @@
+import template from 'views/components/showGroup.html';
+
+angular.module('confRegistrationWebApp')
+  .component('showGroup', {
+    templateUrl: template,
+    bindings: {
+      modalInstance: '<',
+      resolve: '<'
+    },
+    controller: [function(){
```
OzzieOrca: Does this need to be wrapped in an array? I think we have a babel plugin to write the $inject stuff so manually listing what needs to be injected probably isn't needed.

app/scripts/directives/showGroup.js
```diff
@@ -0,0 +1,55 @@
...
+
+        const registration = $ctrl.getRegistration($ctrl.registrationId);
+
+        _.remove($ctrl.visibleRegistrantTypes, function(registrantType) {
+          //remove if type is marked as hidden and a registrant with this type doesn't already exist in the registration
+          return registrantType.hidden && !_.includes(_.map(registration.registrants, 'registrantTypeId'), registrantType.id);
+        });
+      };
+
+      $ctrl.close = function() {
```
OzzieOrca: If you add a binding directly to `dismiss` as I commented above, you could call `$ctrl.resolve.dismiss()` directly in the template instead of having to make a wrapper function.

app/scripts/directives/showGroup.js
```diff
@@ -0,0 +1,55 @@
...
+        if(!type.useLimit){
+          return false;
+        }
+        if(!type.availableSlots){
+          return true;
+        }
+
+        const registration = $ctrl.getRegistration($ctrl.registrationId);
+
+        // subtract currentRegistration's registrants from availableSlots. if there's no slot left, return true
```
OzzieOrca: This comment about subtracting is wrong now... Sorry lol.

Maybe you could say something about returning true when full. I didn't look at the function name at first, though this was returning if there were spots left and thought your logic was backwards but I was wrong.

app/scripts/main.js
```diff
@@ -41,6 +40,7 @@ import 'scripts/directives/pickadate.js';
 import 'scripts/directives/registrationTypeSelect.js';
 import 'scripts/directives/selectOnClick.js';
 import 'scripts/directives/showAnswer.js';
+import 'scripts/directives/showGroup.js';
```
OzzieOrca: Sorry this file is still around. That's annoying. I've had other apps go search for all JS files when I was doing migration to webpack. I sorta remember scanning for all js files didn't work but I can't remember why.

The ideal way to do it is import the JS file you need in the component where it is used. You could do that... and then you wouldn't have to list it here. You could also put the modal component in the same folder as the parent component if that's the only component that uses it. This app has a very flat directory structure which I think makes it hard to find files. I'd prefer files to be grouped by screen and/or feature.

I guess these are just comments for improving the repo in the future. You don't necessarily need to address them in this PR.

app/views/components/showGroup.html
```diff
@@ -1,6 +1,6 @@
 <div class="modal-header">
```
OzzieOrca: I think this template should live next to (in the same folder) as the JS component for it. I know the rest of the app isn't structured this way. I'll let you decide how to structure it. The rest of the templates are here so I guess leaving it here is fine. I do love not having to look in a completely different file path to find templates though.

app/scripts/directives/showGroup.js
```diff
@@ -0,0 +1,55 @@
+import template from 'views/components/showGroup.html';
+
+angular.module('confRegistrationWebApp')
+  .component('showGroup', {
```
OzzieOrca: I did kinda like having modal in the component name and filename. Makes it clear what it is. `showGroupModal` is good.

PR #595, opened 2018-10-01
EVENT-579 Don't display hidden answers

emiliano1:


app/scripts/directives/showAnswer.js
```diff
@@ -21,11 +21,25 @@ angular.module('confRegistrationWebApp')
         }
 
         $scope.getSelectedCheckboxes = function (choices) {
-          return _.keys(_.pickBy(choices, function (val) {
-            return val === true;
+          var blockChoices = $scope.block.content.choices;
+          var posibleChoices = [];
+          for (var i = 0; i < blockChoices.length; i++) {
```
OzzieOrca: Can we use a [`filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) and then a [`map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) here if that makes sense to you? [`reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) would work also if you wanted to do it in one loop. At least a [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) would be better.

PR #596, opened 2018-10-02
EVENT-580 Answer export value

emiliano1:


app/views/modals/choiceOptions.html
```diff
@@ -7,6 +7,12 @@ <h3 translate>Options</h3>
     <label translate>Enter a description to add clarity or more information to registrant.</label>
     <textarea ng-model="choice.desc" class="form-control" style="height: 120px;"></textarea>
   </div>
+  <div class="form-group">
+    <label translate>Export Value</label>
+    <i uib-popover="Enter a value here to replace the answer in a registration export."
```
OzzieOrca: This should probably be marked for translation if we still consider that a priority.

emiliano1: @OzzieOrca  Fixed

PR #597, opened 2018-10-05
EVENT-581 Required questions

emiliano1:


app/views/components/pageRegistration.html
```diff
@@ -15,12 +15,14 @@ <h2 class="page-title border" ng-bind="page.title"></h2>
...
                 type="submit"
-                ng-disabled="savingAnswers"
+                ng-disabled="savingAnswers || !pageIsValid(page.id)"
                 ng-click="goToNext()">
           <span ng-show="!savingAnswers && !nextPage()" translate>Continue</span>
           <span ng-show="!savingAnswers && nextPage()" translate>Next Page</span>
           <span ng-show="savingAnswers" translate>Saving...</span>
         </button>
+
+        <p ng-if="!pageIsValid(page.id)" style="color:red;" class="text-center" translate>Please fill in all required fields.</p>
```
emiliano1: It may be a better idea to move style to a css file.

app/scripts/services/validateRegistrant.js
```diff
@@ -170,6 +170,12 @@ angular.module('confRegistrationWebApp')
               return;
             }
             break;
+          case 'checkboxQuestion':
+            if (_.isEmpty(answer) || _.isEmpty(_.pickBy(answer, function(val){ return val; }))) {
```
OzzieOrca: The default predicate for `_.pickBy` is Lodash's `_.identity` function so you should be able to use it without the second parameter to keep truthy values.
```suggestion
            if (_.isEmpty(answer) || _.isEmpty(_.pickBy(answer)) {
```

app/views/components/pageRegistration.html
OzzieOrca: Looks like there's already a class for it. https://github.com/CruGlobal/conf-registration-web/blob/master/app/styles/variables.scss#L21
Hopefully this works:
```suggestion
        <p ng-if="!pageIsValid(page.id)" class="red text-center" translate>Please fill in all required fields.</p>
```

PR #599, opened 2018-10-23
Fix changing type in admin Add Registrant modal

OzzieOrca:
The data structure for saving the form data could probably be refactored but this seems to work. I tested a little bit but not extensively.


hlbraddock:
I've QA'ed in stage and it works! Let's merge and deploy.

PR #603, opened 2018-12-03
EVENT-603: Support for 'AND' and 'OR' operands for answers rules

bwtoptal:


app/views/components/rule.html
```diff
@@ -57,16 +57,25 @@
...
+            <strong ng-if="choice==='AND'" translate>All rules are met</strong>
+            <strong ng-if="choice==='OR'" translate>Any rules are met</strong>
+          </label>
+        </div>
+      </div>
+    </div>
 
     <div ng-repeat="rule in rules = (getRulesByType(ruleType, answer.value))" style="overflow: auto">
-      <div class="col-md-1 stacked-spacing-col-md" style="line-height: 34px">
-        <span class="pull-left">{{ruleOperandMessage}}</span>
```
OzzieOrca: What was this? Is it still being used? Looks like the rule component takes it in as a binding.

bwtoptal: You're right. We don't need to pass it as a binding. I removed it and replaced the second usage with a static text.

PR #607, opened 2018-12-12
EVENT-584: After preview form page refresh redirect to the start of t…

bwtoptal:


app/scripts/controllers/registration.js
```diff
@@ -31,6 +31,11 @@ angular.module('confRegistrationWebApp')
       return visiblePageArray[_.findIndex(visiblePageArray, {'id': pageId}) + 1];
     };
 
+    // if a page is refreshed in preview mode redirect to the registration welcome page
```
OzzieOrca: Could you put the reason for this in the comment? I had to look up the Jira task to figure out why you were doing this. Hopefully it'll help a future developer.

bwtoptal: Sure, let me know if that's more understandable now.

PR #609, opened 2018-12-20
EVENT-587: Hide question where there are no visible answers (based on…

bwtoptal:
… rules)


app/scripts/services/validateRegistrant.js
```diff
@@ -117,8 +117,23 @@ angular.module('confRegistrationWebApp')
       return !_.includes(block.registrantTypes, registrant.registrantTypeId);
     };
 
+    this.isAnyChoiceVisible = function(block, registrant) {
+      if (block.type !== 'checkboxQuestion' && block.type !== 'selectQuestion'){
+        return true;
+      }
+      for (let i = 0; i < block.content.choices.length; i++) {
```
OzzieOrca: Could use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some.

PR #610, opened 2018-12-21
Event 607 limit registrant types

bwtoptal:


app/scripts/directives/registrationTypeSelect.js
```diff
@@ -6,8 +6,20 @@ angular.module('confRegistrationWebApp')
       templateUrl: template,
       restrict: 'E',
       controller: function ($scope, $rootScope, $location, $routeParams, RegistrationCache, uuid, modalMessage) {
+
         $scope.visibleRegistrantTypes = angular.copy($scope.conference.registrantTypes);
 
+        const findCurrentGroupRegistrantType = function() {
```
OzzieOrca: Could you do something like this?
```js
const findCurrentGroupRegistrantType = (registrants, registrantTypes) => {
  const registrantTypeIds = registrants.map(
    ({ registrantTypeId }) => registrantTypeId
  );
  return registrantTypes.find(
    ({ allowGroupRegistrations, registrantTypeIds }) =>
      allowGroupRegistrations && registrantTypeIds.includes(type.id)
  );
};
```
I think what you've got is a little hard to read with 2 types of loops, having them nested, and the extra conditions.

bwtoptal: I agree, thanks for the tip!

app/views/eventDetails/regTypes.html
```diff
@@ -84,6 +83,31 @@ <h4 translate>Group Registration</h4>
                    popover-append-to-body="true">
                 </i>
               </div>
+              <div ng-if="type.familyStatus == 'ENABLED'" class="bt12">
+                <label translate class="pt12">Associated Registration Types</label>
+                <i class="question-popover fa fa-question-circle"
+                   uib-popover="{{'Select which registrant types are allowed to be added to this registration group and what is the maximum number of them' | translate}}"
```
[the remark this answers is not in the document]
bwtoptal: Changed to what you suggested, thanks ;)

hlbraddock:
@bwtoptal You'll want to **not** display 'Add a Group Member' on the Registration form for Group Registrations when in fact there are no more Registrant Types available to be added.

hlbraddock:
@OzzieOrca Can you review this PR? It's close to ready.

bwtoptal:
Thanks, @hlbraddock! Now this  'Add a Group Member' will be hidden when there are no more registrant types to be added.

app/views/eventDetails/regTypes.html
```diff
@@ -83,6 +83,17 @@ <h4 translate>Group Registration</h4>
                    popover-append-to-body="true">
                 </i>
               </div>
+              <div class="radio">
```
hlbraddock: @bwtoptal @OzzieOrca FYI A slight modification.

PR #619, opened 2019-04-25
Add support for instagram login

hlbraddock:


app/views/modals/loginDialog.html
```diff
@@ -12,6 +12,10 @@ <h3 ng-if="status401" translate>Authorization Required</h3>
       Please sign in with one of the following services:
     </p>
     <div class="row">
+      <div ng-if="instagramLogin" class="{{relayLogin ? 'col-sm-5 col-sm-offset-1' : 'col-sm-6 col-sm-offset-3' }}">
```
hlbraddock: This change alone is not sufficient. Do you have the expertise to quickly and easily modify it? Or is it just something I should tinker with until I get it right? @OzzieOrca

OzzieOrca: You have to adjust the size of the other columns. I could play with it locally if you aren't getting anywhere. Bootstrap has a grid size of 12 so everything in the row has to add up to 12.

hlbraddock: I've updated this commit and this PR is ready for review. Thanks! @OzzieOrca

languages/ert.pot
```diff
@@ -509,10 +509,14 @@ msgid "Create New Event"
 msgstr ""
```
hlbraddock: Is `languages/ert.pot` meant to be auto generated? @OzzieOrca

OzzieOrca: I think running `yarn angular-gettext-extract` generates it. https://github.com/CruGlobal/conf-registration-web/blob/master/package.json#L76

Do we have any live translations? I don't see a whole lot of tooling for translations. I've added OneSky upload/download scripts for MissionHub web and mobile that run on Travis deploys.

app/scripts/app.config.js
```diff
@@ -30,15 +30,15 @@ angular.module('confRegistrationWebApp')
       },
       vars: {
         development: {
-          apiUrl: 'https://api.stage.eventregistrationtool.com/eventhub-api/rest/',
+          apiUrl: 'http://localhost:8080/eventhub-api/rest/',
```
OzzieOrca: Did you mean to commit this?

languages/ert.pot
hlbraddock: ERT isn't officially internationalized, so I don't know why we're doing this anyway.

app/scripts/app.config.js
```diff
@@ -30,15 +30,15 @@ angular.module('confRegistrationWebApp')
...
           tsysEnvironment: 'staging'
         },
         staging: {
-          apiUrl: 'https://api.stage.eventregistrationtool.com/eventhub-api/rest/',
+          apiUrl: 'http://localhost:8080/eventhub-api/rest/',
           tsysEnvironment: 'production'
         },
         production: {
-          apiUrl: 'https://api.eventregistrationtool.com/eventhub-api/rest/',
+          apiUrl: 'http://localhost:8080/eventhub-api/rest/',
```
OzzieOrca: We can't go to production with this. Changing the `development` one should have been enough if you were testing on localhost. This environment library just does string matching with the domain in the address bar.

hlbraddock:
https://github.com/CruGlobal/conf-registration-web/blob/instagram-login/app/img/signin-instagram.jpeg and https://github.com/CruGlobal/conf-registration-web/blob/instagram-login/app/img/signin-instagram-grey.jpeg are actually identical images. I didn't grey one out. And I don't know where it's used.

OzzieOrca:
Was that a question? If you added the grey version and aren't using it, let's delete it. There may have been some old unused versions (maybe the grey ones) of the other sign in logos running around.

app/scripts/app.config.js
hlbraddock: My mistake. I keep inadvertently including that in the commit. Resolved.

PR #621, opened 2019-04-29
Added debounce to question input fields to delay function firing

TheNoodleMoose:


app/scripts/services/validateRegistrant.js
```diff
@@ -173,7 +173,7 @@ angular
       );
     };
 
-    var blockInRegistrantType = function(block, registrant) {
+    this.blockInRegistrantType = function(block, registrant) {
```
OzzieOrca: How come you made this an instance method? I think it's fine either way but usually you expose it on the class if  you need it in the template or want to share it externally. Just wondering the motivation.

app/scripts/services/validateRegistrant.js
```diff
@@ -185,7 +185,8 @@ angular
       ) {
         return true;
       }
-      for (let i = 0; i < block.content.choices.length; i++) {
+
+      for (let i = 0, len = block.content.choices.length; i < len; i++) {
```
OzzieOrca: This is a little weird looking. Guess it might slightly help performance. Keep it.

I would probably replace this whole thing with an [`Array.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) but that may be less performant than a `for` loop. Usually I pick the type of loop to make the code easier to read/maintain and not necessarily performance. Usually any performance difference is trivial, especially if you have a small amount of data but not always...

app/scripts/services/validateRegistrant.js
```diff
@@ -194,15 +195,16 @@ angular
     };
 
     this.blockVisible = function(block, registrant, isAdmin) {
-      var visible =
+      let visible =
```
OzzieOrca: This could be `const` instead.

app/scripts/services/validateRegistrant.js
```diff
@@ -233,7 +235,6 @@ angular
       var blocks = page
         ? _.find(conference.registrationPages, { id: page }).blocks
         : _.flatten(_.map(conference.registrationPages, 'blocks'));
-
       var that = this;
       angular.forEach(blocks, function(block) {
```
OzzieOrca: If you make `function(block) { ... }` an arrow function you don't have to do the `that = this` hack.

I know you didn't write this code... You can leave it...

app/scripts/services/validateRegistrant.js
TheNoodleMoose: I think I did this when I was testing a different solution that required exposing blockInRegistrantType, but it did not end up working. I can change it back because it isn't necessary any more.

TheNoodleMoose:
I originally tested with around 200ms, 300ms and that was not enough. I think I tried 500ms and it was better but 1000ms seemed like the perfect value. I can test to see what 750ms looks like? I agree that this seems more like a bandaid solution, while fixing the underlying logic is the true fix. It seems that the issue is that logic just calls so many other expensive functions for every single dropdown option on this conference. The blockVisible function seems to be the culprit, as it is what is checked for each block to see if it should be shown or not, and it is what calls every expense function such as blockVisibleRuleCheck, isAnyChoiceVisible, etc.

OzzieOrca:
Ok good to know you tested those. We can deploy 1000ms but I'd rather not leave that in production long term. I guess you could test 750ms and see if that's enough to make the big conference usable.

Ya this code is a mess...

TheNoodleMoose:
I tried 750ms but I think 1000ms is the number to use for now. Yeah, the logic behind it makes sense but the way it is executed is a mess.I will continue to look into a way to optimize the logic. If you have any suggestions they would be greatly appreciated.

app/scripts/services/validateRegistrant.js
OzzieOrca: This could be a `const`.

app/scripts/services/validateRegistrant.js
```diff
@@ -194,49 +209,49 @@ angular
...
+
       conference = angular.copy(conference);
-      var blocks = page
+      const blocks = page
         ? _.find(conference.registrationPages, { id: page }).blocks
         : _.flatten(_.map(conference.registrationPages, 'blocks'));
-      var that = this;
-      angular.forEach(blocks, function(block) {
+
+      _.map(blocks, block => {
```
OzzieOrca: I love `map` but if you aren't using the return value, `_.forEach` is probably better.

app/views/registration.html
```diff
@@ -148,7 +153,9 @@ <h2 class="page-title border details-heading" translate>
         </div>
         <div class="col-sm-9 col-xs-6">
           <a
-            ng-href="{{registerMode}}/{{conference.id}}/page/{{conference.registrationPages[0].id}}?reg={{currentRegistration.registrants[0].id}}"
+            ng-href="{{ registerMode }}/{{ conference.id }}/page/{{
+              conference.registrationPages[0].id
+            }}?reg={{ currentRegistration.registrants[0].id }}"
```
OzzieOrca: Are you changing the formatting on this HTML or did Prettier somehow not run on these files initially? These changes are fine.

app/scripts/directives/blockRegistration.js
```diff
@@ -77,24 +77,20 @@ angular
...
+            ) {
+              return;
+            }
 
-              RegistrationCache.updateCurrent(
-                $scope.conference.id,
-                $scope.currentRegistration,
-              );
-            },
-            true,
```
OzzieOrca: Have you tested this? Removing this disables deep equality checks right? If we don't need it, that's great, just seems like it might miss updates if `answer` is an object or array.

app/scripts/services/validateRegistrant.js
TheNoodleMoose: Yeah, you're right. I will convert this back

app/views/registration.html
TheNoodleMoose: Prettier seems to not have ran on some files here and there which is strange.

app/scripts/directives/blockRegistration.js
TheNoodleMoose: I read an article about improving angularJS performance and this was a suggestion. Everything seems to run fine but no performance boost was gained so I will switch it back just in case.

PR #623, opened 2019-05-16
Added link to React Dev Env

TheNoodleMoose:
Looks like prettier did not format this file the first time. My changes starts on line 283 and end on line 289.

### DIFF
--- a/app/index.ejs
+++ b/app/index.ejs
@@ -280,12 +280,12 @@
                   >Local Dev Environment</a
                 >
               </li>
-              <% } %> <% if(!htmlWebpackPlugin.options.prod) { %>
               <li>
                 <a ng-href="http://localhost:3000/auth/{{ crsToken }}"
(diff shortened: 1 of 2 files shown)


app/index.ejs
```diff
@@ -1,201 +1,346 @@
...
+                >
+              </li>
+              <li><a href="help/" translate>Help</a></li>
+              <% if(!htmlWebpackPlugin.options.prod) { %>
+              <li>
+                <a ng-href="http://localhost:9000/auth/{{ crsToken }}"
+                  >Local Dev Environment</a
+                >
+              </li>
+              <% } %> <% if(!htmlWebpackPlugin.options.prod) { %>
```
OzzieOrca: I think you could just keep the same `if` going for both links.

TheNoodleMoose: Yeah you're right, changed it :+1:

PR #624, opened 2019-05-21
Upgraded Webpack to address security alert

TheNoodleMoose:


package.json
```diff
@@ -34,22 +34,22 @@
     "css-loader": "^0.28.4",
     "eslint": "^4.0.0",
     "eslint-config-angular": "^0.5.0",
-    "eslint-loader": "^1.8.0",
+    "eslint-loader": "^2.0.0",
     "eslint-plugin-angular": "^2.4.2",
     "eslint-plugin-jasmine": "^2.3.0",
-    "extract-text-webpack-plugin": "^2.1.0",
+    "extract-text-webpack-plugin": "^4.0.0-beta.0",
```
OzzieOrca: This still needed? Looks like you stopped using it in the webpack config.

package.json
```diff
@@ -34,22 +34,22 @@
...
-    "html-webpack-plugin": "^2.28.0",
+    "html-webpack-plugin": "^3.2.0",
     "husky": "^2.1.0",
     "image-webpack-loader": "^3.3.1",
     "inline-manifest-webpack-plugin": "^3.0.1",
     "istanbul-instrumenter-loader": "^2.0.0",
     "jasmine-core": "^2.5.2",
     "jasmine-promise-matchers": "^2.2.0",
-    "json-loader": "^0.5.4",
+    "json-loader": "^0.5.7",
```
OzzieOrca: You can probably toss this too.

webpack.config.js
```diff
@@ -29,8 +30,7 @@ const htmlMinDefaults = {
   removeStyleTypeAttributes: true,
 };
 
-module.exports = env => {
-  env = env || {};
+module.exports = (env = {}) => {
```
OzzieOrca: 👍

webpack.config.js
```diff
@@ -50,8 +50,9 @@ module.exports = env => {
           jQuery: 'jquery',
           'window.jQuery': 'jquery',
         }),
-        new ExtractTextPlugin({
-          filename: '[name].[contenthash].min.css',
+        new MiniCssExtractPlugin({
+          filename: '[name].[contenthash].scss',
```
OzzieOrca: Is this needed? An object with 2 identical keys will use the second one. Also I don't think we want to output `scss`.

webpack.config.js
```diff
@@ -159,23 +160,22 @@ module.exports = env => {
...
+            {
+              loader: 'css-loader',
+              options: {
+                sourceMap: true,
               },
-              {
-                loader: 'sass-loader',
-                options: {
-                  sourceMap: true,
-                  precision: 10, // fixes line height issue with bootstrap button addons
```
OzzieOrca: Can you keep this comment around? Just cuz it's so random and I don't want someone accidentally deleting the precision config. Or if it's no longer needed, toss the config?

But seems like it's still needed. https://github.com/twbs/bootstrap-sass#sass-number-precision. You could add that link to give it more context. Might also be able to drop the precision from 10 to 8...

PR #625, opened 2019-05-23
Fixed issue where an empty radio question would cause page to be rend…

TheNoodleMoose:
…ered as non visible


TheNoodleMoose:
> This is the admin forgot to add choices or the user didn't choose any answers?

This is when the admin forgot to add choices.

OzzieOrca:
Hmm ok. Might be better to handle it on the admin side so a question with no answers isn't saved. But I don't really know how. This is fine.

PR #626, opened 2019-06-25
Use Adobe Launch on staging

OzzieOrca:
Prettier reformatted the file. But you can review it ignoring whitespace changes. I also merged it into staging.
Screen Shot 2019-06-25 at 12 07 39 PM

https://secure.helpscout.net/conversation/888002030/336283/?folderId=610613

### DIFF
--- a/app/index.ejs
+++ b/app/index.ejs
@@ -1,201 +1,372 @@
-<!doctype html>
+<!DOCTYPE html>
 <html ng-app="confRegistrationWebApp">
-<head>
-  <meta charset="UTF-8">
-  <meta http-equiv="X-UA-Compatible" content="IE=edge">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
-  <base href="/">
+  <head>
+    <meta charset="UTF-8" />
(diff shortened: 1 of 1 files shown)


app/index.ejs
```diff
@@ -1,201 +1,372 @@
...
+      );
 
-<script type="text/javascript">window._satellite && _satellite.pageBottom();</script>
+      ga('create', 'UA-325725-55', 'auto');
+    </script>
 
-<%=htmlWebpackPlugin.files.webpackManifest%>
-</body>
+    <script type="text/javascript">
+      window._satellite && _satellite.pageBottom();
```
OzzieOrca: @brenton05, this doesn't need to be called anymore with Launch right? I put in the if statement that only runs in prod. We'll remove it when we add the prod scripts.

brenton05: referring to line 197? If so, it does not if we are using the launch script I gave you but again, only on stage. Prod will still need the pageBottom call.

OzzieOrca: K this code should be good then. It'll only call pageBottom on prod.

PR #627, opened 2019-06-26
Addressed issues with webpack upgrade causing build to fail

TheNoodleMoose:
The first issue that was causing the build fail was: 
```js
      new webpack.optimize.CommonsChunkPlugin({	
              name: 'vendor',	
              minChunks: function(module) {	
                // This prevents stylesheet resources with the .css or .scss extension	
                // from being moved from their original chunk to the vendor chunk	
                if (	
                  module.resource &&	
                  /^.*\.(css|scss)$/.test(module.resource)	
                ) {	
                  return false;	
                }	
                return (	
                  module.context &&	
                  module.context.indexOf('node_modules') !== -1	
                );	
              },	
            }),	
      new webpack.optimize.CommonsChunkPlugin({	
              name: 'manifest',	
              minChunks: Infinity,	
            }),
```
CommonChunkPlugin has been deprecated as of webpack v4, so I replaced it with SplitChunks Plugin(https://webpack.js.org/plugins/split-chunks-plugin/). As far as naming conventions, if you have a better suggestion I'd appreciate it.

After that I upgraded _html-webpack-exclude-assets-plugin_ to the version that is compatible with webpack 4. Then I switched _favicons-webpack-plugin_ to a webpack 4 compatible and maintained package, _webapp-webpack-plugin_. This replacement does the samething as the old one does. Also replaced _webpack-manifest-plugin_ with _webpack-inline-manifest-plugin_. Similar situation, new package is well maintained and works with webpack 4. 

And finally I just added the link to the local React Dev environment back in. Running yarn build and yarn start both had successful compiles.


webpack.config.js
```diff
@@ -43,6 +43,31 @@ module.exports = (env = {}) => {
         info.resourcePath.replace(/^\.\//, ''),
       crossOriginLoading: 'anonymous',
     },
+    optimization: {
+      splitChunks: {
+        cacheGroups: {
+          commons: {
+            name: 'vendor',
+            test: /[\\/]node_modules[\\/]/,
+            chunks: function(module) {
```
OzzieOrca: This doesn't seem to be splitting chunks right. Here's the current output of running `yarn build:analyze` on this branch:
Screen Shot 2019-06-26 at 10 28 18 AM

I changed the config to match https://github.com/CruGlobal/missionhub-web/blob/master/webpack.config.js#L48-L58 and got this:
Screen Shot 2019-06-26 at 10 28 25 AM

I couldn't instantly find good docs on what `chunks: 'initial',` does and don't remember why I did that before... Webpack is always fun :)

TheNoodleMoose: Should I leave the commons2, manifest, to have a minChunks of Infinity still? If not I'll push up the change with the new config of initial :slightly_smiling_face: I did not know yarn build:analyze was a thing but that is super cool!

OzzieOrca: I don't really know what that does. If there's a reason to do it that should be good. But simpler config is always nice :)

TheNoodleMoose: Yeah I'm not sure either. Simple is definitely better. Webpack has been a learning experience for sure :joy:

webpack.config.js
```diff
@@ -43,6 +43,22 @@ module.exports = (env = {}) => {
...
     },
+    optimization: {
+      splitChunks: {
+        cacheGroups: {
+          commons: {
+            name: 'vendor',
+            test: /[\\/]node_modules[\\/]/,
+            chunks: 'initial',
+          },
+          commons2: {
```
OzzieOrca: So what does `commons2` do? I ran analyze with and without it and don't see any difference...

TheNoodleMoose: Yeah, you're right. It does not seem to affect anything either way. I had brought it over initially because there was a CommonChunksPlugin for it.  I'll go ahead and remove it just for simplicity sake. :+1:

TheNoodleMoose:
When I switch it to be merged to staging I get popup saying: 

> _Some commits from the old base branch may be removed from the timeline, and old review comments may become outdated._

Is that okay? I just don't want to do something incorrect lol

OzzieOrca:
Oh if you do it on github and merge it, it'll close your PR and you'll have to make another PR to go to master. Usually I just merge and push to staging locally. You can reset staging to master first since it's got some clutter from your old link thing and my analytics commit.

TheNoodleMoose:
Just pushed to staging and it successfully built  :+1:

app/index.ejs
```diff
@@ -1,201 +1,379 @@
...
+              },
+            ],
+          },
+        ],
+      };
+    </script>
+    <% if(htmlWebpackPlugin.options.prod) { %>
+    <script src="//assets.adobedtm.com/3202ba9b02b459ee20779cfcd8e79eaf266be170/satelliteLib-928beaae1bce68f8836cc63dd18a9951aa3a1f4e.js"></script>
+    <% } else { %>
+    <script src="//assets.adobedtm.com/3202ba9b02b459ee20779cfcd8e79eaf266be170/satelliteLib-928beaae1bce68f8836cc63dd18a9951aa3a1f4e-staging.js"></script>
```
OzzieOrca: This should have been deleted. We're replacing the DTM/satellite script with the Adobe Launch one. I think your merge kept it around.

Also you deployed everything on staging to master, which in this case was just my branch https://github.com/CruGlobal/conf-registration-web/pull/626. That won't hurt anything being on prod but usually you don't want to do that cuz there could be stuff that isn't ready/is untested.

I probably could have explained it better but usually you leave the branch and the PR alone and just merge  things directly to the staging branch that you need to test. If the staging branch gets to messy I usually reset it to master unless other developers are depending on things being there. And then merge the PRs to master individually.

TheNoodleMoose: Oh I apologize for this. I have never done a reset to master, is there a simple way too reset to what is currently on master via Github? Or do you do it locally, and then push up to staging?

OzzieOrca: Just locally. I mean you could make a PR with the base set as staging to merge things into staging but that's more work.

OzzieOrca: Everything should be good and up to date now

TheNoodleMoose: Sweet, thanks for the help Scotty!  :slightly_smiling_face:

PR #629, opened 2019-07-02
Use Adobe Launch on production

OzzieOrca:
No one gave me a new script url. I'm just assuming it's the stage one without the `-development`.


TheNoodleMoose:
@OzzieOrca Do you think this pr is still relevant or can it be closed since we are migrating to google tag manager now?

OzzieOrca:
Nope, closing it. Idk why it never got merged. I don't remember if I was waiting for approval or something. I think you're the expert on what's happening in the analytics world, so do what you need to.

PR #631, opened 2019-09-13
EVENT-638: Registration completion redirect links on a per registrant type - Web

katefree:
See https://jira.cru.org/browse/EVENT-638


app/scripts/controllers/eventDetails.js
```diff
@@ -275,22 +275,22 @@ angular
...
-          if (
-            !httpPattern.test($scope.conference.registrationCompleteRedirect)
-          ) {
-            $scope.conference.registrationCompleteRedirect =
-              'http://' + $scope.conference.registrationCompleteRedirect;
+      angular.forEach($scope.conference.registrantTypes, function(type) {
+        if (!_.isEmpty(type.registrationCompleteRedirect)) {
+          urlPattern.lastIndex = 0;
+          httpPattern.lastIndex = 0;
+          if (!urlPattern.test(type.registrationCompleteRedirect)) {
```
TheNoodleMoose: I think here you could do something like to clean up the if statements, not a big deal though: 
```suggestion
!urlPattern.test(type.registrationCompleteRedirect)
            ? validationErrors.push(
                'Please enter a valid completion redirect website.',
              )
            : !httpPattern.test(type.registrationCompleteRedirect)
            ? (type.registrationCompleteRedirect =
                'http://' + type.registrationCompleteRedirect)
            : null;
```

app/scripts/controllers/reviewRegistration.js
```diff
@@ -121,14 +118,28 @@ angular
       });
     }
 
+    function getRegistrantType(typeId) {
```
TheNoodleMoose: I think we could .find instead of using lodash here: 
```suggestion 
function getRegistrantType(typeId) {
      return conference.registrantTypes.find(type => type === typeId);
    }
```

app/scripts/controllers/reviewRegistration.js
```diff
@@ -121,14 +118,28 @@ angular
...
+      return _.find(conference.registrantTypes, {
+        id: typeId,
+      });
+    }
+
+    function primaryRegType(currentRegistration) {
```
TheNoodleMoose: Similar thing here too: 

```suggestion
    function primaryRegType(currentRegistration) {
      let primaryRegistrant = currentRegistration.registrants.find(
        registrant => registrant.id === currentRegistration.primaryRegistrantId,
      );
      return getRegistrantType(primaryRegistrant.registrantTypeId);
    }
```

app/scripts/controllers/reviewRegistration.js
```diff
@@ -121,14 +118,28 @@ angular
...
+      return getRegistrantType(primaryRegistrant.registrantTypeId);
+    }
+
     // Navigate to the correct page after completing a registration
-    function navigateToPostRegistrationPage() {
-      if ($scope.regType.registrationCompleteRedirect) {
-        $window.location.href = $scope.regType.registrationCompleteRedirect;
+    $scope.navigateToPostRegistrationPage = function() {
+      let regType = primaryRegType(currentRegistration);
+      if (regType.registrationCompleteRedirect) {
```
TheNoodleMoose: Could probably do ternary here as well:
```suggestion
regType.registrationCompleteRedirect
        ? ($window.location.href = regType.registrationCompleteRedirect)
        : $route.reload();
```

app/scripts/controllers/reviewRegistration.js
katefree: I've just tried to replace it with .find but it failed on CI, looks like there is no such function there:
> TypeError: undefined is not a function (evaluating 'conference.registrantTypes.find(function (type) {
Should add a polyfill for this function? I reverted it to use lodash for now.

app/scripts/controllers/reviewRegistration.js
katefree: The same as for https://github.com/CruGlobal/conf-registration-web/pull/631/files#r324827925

app/scripts/controllers/eventDetails.js
katefree: done

app/scripts/controllers/reviewRegistration.js
katefree: done

app/scripts/controllers/reviewRegistration.js
TheNoodleMoose: Hmm, .filter is actually what I should have put but the lodash should be fine honestly. We use it everywhere else.

PR #634, opened 2019-10-15
EVENT-642: Generate Cash and check deposit report Web

katefree:


app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
+angular
+  .module('confRegistrationWebApp')
+  .service('paymentReportService', function PaymentReportService(
+    $cacheFactory,
+    $rootScope,
+    $http,
+    $q,
+  ) {
+    const path = function(id) {
+      return 'conferences/' + (id || '') + '/payments/report';
```
TheNoodleMoose: Is there any instance where we would generate this path and not have the id?

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+      let queryParamForExcludedPayments = this.queryParamForExcludedPayments(
+        excludedIdsMap,
+      );
+      $http
+        .post(
+          path(conferenceId) +
+            '/lock' +
+            (queryParamForExcludedPayments
+              ? '?' + queryParamForExcludedPayments
+              : ''),
```
TheNoodleMoose: Couldn't you write this with template literals?
```suggestion 
`${path(conferenceId)/lock${queryParamForExcludedPayments ? `?${queryParamForEcludedPayments}` : ''}}`,
```
Not that this much cleaner but just a nitpicky/personal preference thing.

test/spec/controllers/eventRegistrations.spec.js
```diff
@@ -104,4 +104,15 @@ describe('Controller: eventRegistrations', function() {
     scope.registerUser();
     expect(openModal).toHaveBeenCalled();
   });
+
+  it('refreshRegistrations should assign reported attribute to registarnts', function() {
```
TheNoodleMoose: Type in test description
```suggestion
  it('refreshRegistrations should assign reported attribute to registrants', () => {
```

app/scripts/services/paymentReportService.js
katefree: It is always run in the context of conference. This path can be used for both viewing of existing reports and generation of new one for the conference based on currently existing payments

app/scripts/services/paymentReportService.js
katefree: fixed

app/scripts/services/paymentReportService.js
OzzieOrca: I don't think you really answered the question. If the id can't be undefined, then you shouldn't need the `|| ''`. If it can be undefined, `'conferences//payments/report'` doesn't really make sense. Also you could use template literals.

app/views/paymentReport.html
```diff
@@ -0,0 +1,165 @@
...
+              <td>
+                <input
+                  type="checkbox"
+                  ng-model="r.included"
+                  ng-change="refreshExcludedIds(r.paymentId)"
+                  ng-disabled="queryParameters.currentReportId"
+                />
+              </td>
+            </tr>
+            <tr ng-repeat-end class="rowdetails on"></tr>
```
OzzieOrca: What does this do? Just add spacing below or something? https://github.com/CruGlobal/conf-registration-web/blob/347660a9b33ea6f9d5342a26e27edc24f04e7133/app/styles/event-admin.scss#L216 is the only definition of it and I don't see any `well`s around.
https://github.com/CruGlobal/conf-registration-web/blob/347660a9b33ea6f9d5342a26e27edc24f04e7133/app/views/eventRegistrations.html#L573 had it to display a well with details.

app/scripts/controllers/eventRegistrations.js
```diff
@@ -175,7 +175,13 @@ angular
...
           $scope.registrations = data.registrations;
           $scope.registrants = _.flatten(
-            _.map(data.registrations, 'registrants'),
+            _.map(data.registrations, r => {
+              let regs = r.registrants;
+              for (const reg of regs) {
+                reg.reported = r.reported;
+              }
+              return regs;
+            }),
```
OzzieOrca: Can we be consistent on loop types here? My preference would be vanilla JS map/filter/reduce.

```js
$scope.registrants = data.registrations.flatMap(registration =>
  registration.registrants.map(registrant => ({
    ...registrant,
    reported: registration.reported
  }))
);
```

OzzieOrca: I would also prefer not to clone and denormalize data as it tends to lead to bugs and state inconsistencies. Idk if there's a good way to look this up though... You could do `$scope.registrations[registrant.registrationId].reported`... As you don't use this very often in the template, I think looking it up when you need it would be doable.

app/scripts/controllers/paymentReport.js
```diff
@@ -0,0 +1,119 @@
...
+        'conferences/' +
+        $scope.report.conferenceId +
+        '/payments/report/export' +
+        currentReportId +
+        '?Authorization=' +
+        $scope.authToken +
+        (queryParamForExcludedPayments
+          ? '&' + queryParamForExcludedPayments
+          : '')
+      );
```
OzzieOrca: I think template literals would make this easier to read.

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+    $rootScope,
+    $http,
+    $q,
+  ) {
+    const path = function(id) {
+      return 'conferences/' + (id || '') + '/payments/report';
+    };
+
+    function getReportData(url) {
+      const defer = $q.defer();
```
OzzieOrca: Can we not use `defer`? You should just be able to return the promise returned by `$http.get` directly.

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+      $rootScope.loadingMsg = 'Loading Details';
+      $http
+        .get(url)
+        .then(function(response) {
+          var report = response.data;
+          defer.resolve(report);
```
OzzieOrca:
```suggestion
          return response.data;
```

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+        })
+        .catch(function(error) {
+          defer.reject(error);
+        })
```
OzzieOrca: If you wanted to rethrow you could do
```js
.catch(function(error) {
  throw error;
})
```
But you aren't doing anything with the error or transforming it in any way so you can just delete the whole `catch` block and the error will propagate.

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+        .get(path(conferenceId) + '/' + report, {
+          params: queryParameters,
+        })
+        .then(function(response) {
+          $rootScope.loadingMsg = '';
+          defer.resolve(response.data);
+        })
+        .catch(function() {
+          $rootScope.loadingMsg = '';
+          defer.reject();
```
OzzieOrca:
```suggestion
          throw error;
```

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+      let report = queryParameters.currentReportId
+        ? queryParameters.currentReportId
+        : 'new';
+      $http
+        .get(path(conferenceId) + '/' + report, {
+          params: queryParameters,
+        })
+        .then(function(response) {
+          $rootScope.loadingMsg = '';
+          defer.resolve(response.data);
```
OzzieOrca:
```suggestion
          return response.data;
```

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+      return defer.promise;
+    }
+
+    this.getAll = function(id) {
+      var allReportsUrl = path(id);
+      return getReportData(allReportsUrl);
+    };
+
+    this.getReport = function(conferenceId, queryParameters) {
+      var defer = $q.defer();
```
OzzieOrca: Return the $http promise instead of using defer.

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+
+    this.queryParamForExcludedPayments = function(excludedIdsMap) {
+      let excludedIds = this.collectExcludedIds(excludedIdsMap);
+      return excludedIds.length > 0
+        ? 'excludedPaymentsIds=' + excludedIds.join('&excludedPaymentsIds=')
+        : '';
+    };
+
+    this.lockReport = function(conferenceId, excludedIdsMap) {
+      var defer = $q.defer();
```
OzzieOrca: Return the $http promise.

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+          `${path(conferenceId)}/lock${
+            queryParamForExcludedPayments
+              ? '?' + queryParamForExcludedPayments
+              : ''
+          }`,
+          {},
+        )
+        .then(function(response) {
+          $rootScope.loadingMsg = '';
+          defer.resolve(response.data);
```
OzzieOrca:
```suggestion
          return response.data;
```

app/scripts/services/paymentReportService.js
```diff
@@ -0,0 +1,95 @@
...
+        })
+        .catch(function() {
+          $rootScope.loadingMsg = '';
+          defer.reject();
```
OzzieOrca:
```suggestion
          throw error;
```

app/scripts/services/paymentReportService.js
katefree: The or is removed now

app/views/paymentReport.html
katefree: I just removed it, it is definetely not needed

app/scripts/controllers/eventRegistrations.js
katefree: Reverted this change, added `isRegistrantReported` function instead

app/scripts/controllers/paymentReport.js
katefree: fixed

app/scripts/services/paymentReportService.js
katefree: fixed

app/scripts/services/paymentReportService.js
katefree: fixed

app/scripts/services/paymentReportService.js
katefree: done

app/scripts/services/paymentReportService.js
katefree: removed catch block

app/scripts/services/paymentReportService.js
katefree: done

app/scripts/services/paymentReportService.js
katefree: done

app/scripts/services/paymentReportService.js
katefree: done

app/scripts/services/paymentReportService.js
katefree: done

app/scripts/services/paymentReportService.js
katefree: done

app/views/paymentReport.html
```diff
@@ -0,0 +1,165 @@
...
+              <td>
+                <input
+                  type="checkbox"
+                  ng-model="r.included"
+                  ng-change="refreshExcludedIds(r.paymentId)"
+                  ng-disabled="queryParameters.currentReportId"
+                />
+              </td>
+            </tr>
+            <tr ng-repeat-end></tr>
```
OzzieOrca: I guess part of my question was whether you need the `<tr ng-repeat-end></tr>` at all? You could just use `ng-repeat` above instead of `ng-repeat-start` if that's the case.

app/scripts/controllers/eventRegistrations.js
```diff
@@ -314,6 +321,7 @@ angular
             var index = _.findIndex($scope.registrants, {
               id: registrantData.id,
             });
+            registrantData.reported = $scope.registrants[index].reported;
```
OzzieOrca: This still needed?

app/scripts/app.config.js
```diff
@@ -219,6 +220,36 @@ angular
           },
         },
       })
+      .when('/paymentReport/:conferenceId', {
+        title: gettext('Payment Report Preview'),
+        templateUrl: paymentReportTemplate,
+        controller: 'paymentReportCtrl',
```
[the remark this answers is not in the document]
katefree: Would it mean that we need the new parent component with its own routing?

app/scripts/controllers/eventRegistrations.js
```diff
@@ -186,6 +186,13 @@ angular
       );
     };
 
+    $scope.isRegistrantReported = function(registrant) {
+      let t = $scope.registrations.find(r =>
+        r.registrants.some(rr => rr.id === registrant.id),
+      );
+      return t && t.reported;
+    };
```
OzzieOrca: I like this better :)

I think it can be a little simpler:
```js
$scope.isRegistrantReported = function(registrant) {
  const registration = $scope.registrations.find(r => r.id === registrant.registrationId);
  return registration && registration.reported;
};
```

app/scripts/app.config.js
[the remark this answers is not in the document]
OzzieOrca: You can leave it since it would be a little bit of a rewrite. Since this app is very dependant on shared scopes (which is so messy) it might be hard to start isolating things but it would be a lot nicer.

But ya components are nice since they enforce some of the best practices from directives, keep everything grouped together, and have isolate scope. https://docs.angularjs.org/guide/component

Here are some [example components from another project](https://github.com/CruGlobal/missionhub-web/tree/master/app/assets/javascripts/angular/components) and the [router config](https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/javascripts/angular/missionhubApp.routes.js#L245).

PR #647, opened 2019-10-21
EVENT-644: currency symbol based on user locale 

katefree:


app/views/eventDetails/regTypes.html
```diff
@@ -280,8 +280,12 @@ <h4 translate>Payment</h4>
                 <label translate>Minimum Payment to Register</label>
                 <div class="input-group">
                   <span class="input-group-addon"
-                    ><i class="fa fa-usd"></i
-                  ></span>
+                    ><i class="fa"
```
OzzieOrca: Is this still using font awesome? If the `fa` style is exactly what you want, you can keep it. Just seems a little odd to have font awesome html and then not use font awesome for the icon.

app/views/eventDetails/regTypes.html
```diff
@@ -280,8 +280,12 @@ <h4 translate>Payment</h4>
...
+                      ><span
+                        >&nbsp;{{conference.currency.shortSymbol}}&nbsp;</span
```
OzzieOrca: You could probably replace `&nbsp;` with padding.

app/views/eventDetails/regTypes.html
```diff
@@ -356,8 +360,12 @@ <h4 translate>Early Registration Discount(s)</h4>
                   <label translate>Discount</label>
                   <div class="input-group">
                     <span class="input-group-addon"
-                      ><i class="fa fa-usd"></i
-                    ></span>
+                      ><i class="fa"
+                        ><span
+                          >&nbsp;{{conference.currency.shortSymbol}}&nbsp;</span
```
OzzieOrca: Same here.

app/views/modals/choiceOptions.html
```diff
@@ -31,7 +31,7 @@ <h3 translate>Options</h3>
     <label translate>Additional Cost</label>
     <div class="input-group">
       <span class="input-group-addon"
-        ><span>&nbsp;{{conference.currency.shortSymbol}}&nbsp;</span></span
+        ><span>&nbsp;{{currency}}&nbsp;</span></span
```
OzzieOrca: Same here.

app/views/modals/paymentsModal.html
```diff
@@ -386,8 +386,12 @@ <h4>
                       >
                         <div class="input-group">
                           <span class="input-group-addon"
-                            ><i class="fa fa-usd"></i
-                          ></span>
+                            ><i class="fa"
+                              ><span
+                                >&nbsp;{{conference.currency.shortSymbol}}&nbsp;</span
```
OzzieOrca: Same here.

app/views/eventDetails/regTypes.html
katefree: I have tried just to remove it and it become italic. These fa class unfortunately exists also in many other places so maybe it worth refactor it in the separate pull request, please let me know what you think

app/views/eventDetails/regTypes.html
katefree: fixed

app/views/eventDetails/regTypes.html
katefree: fixed

app/views/modals/choiceOptions.html
katefree: fixed

app/views/modals/paymentsModal.html
katefree: fixed

app/views/eventDetails/regTypes.html
OzzieOrca: Nothing's forcing you to use the `<i>` tag. It's just the element font awesome decided to use for their icons.

PR #651, opened 2019-11-01
EVENT-637: For Cru events: Require ministry and event type

katefree:
This PR attempts to eliminate `Transaction Rolled Back` errors, as documented in https://jira.cru.org/browse/EVENT-655.

### DIFF
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -213,7 +213,10 @@ angular
 
     $scope.saveEvent = function() {
       //validation check
+      const eventInformationPageHint =
+        "<strong>*</strong>You can provide this information on the 'Event Information' page.";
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/eventDetails.js
```diff
@@ -414,6 +414,26 @@ angular
...
+      }
+
+      if (
+        $scope.conference.cruEvent &&
+        $scope.conference.ministry &&
+        $scope.getSubministries().length !== 0 &&
+        !$scope.conference.subministry
+      ) {
+        validationErrors.push(
+          'Please enter Which Strategy is hosting the event if applicable.',
```
OzzieOrca:
```suggestion
          'Please enter which Strategy is hosting the event if applicable.',
```

PR #662, opened 2020-01-09
LocalizedCurrencyFilter_Fix

TheNoodleMoose:
This is another fix for that same help scout ticket. Basically it seems some conferences didn't have their currencyCode set to the default 'USD' by the API. Lee is eventually going to make a check that will set it no matter what API Side, but for now to help the user we are doing it client side.

### DIFF
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -79,11 +79,18 @@ angular
     $scope.conference = angular.copy(conference);
     $scope.currencies = currencies;
 
-    // If by chance the conference currency code doesn't exist
+    // If by chance the conference currency or currency code doesn't exist
     // Set the default to USD.
-    if (!$scope.conference.currency.currencyCode) {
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/eventDetails.js
```diff
@@ -79,11 +79,18 @@ angular
...
 
-    // If by chance the conference currency code doesn't exist
+    // If by chance the conference currency or currency code doesn't exist
     // Set the default to USD.
-    if (!$scope.conference.currency.currencyCode) {
-      $scope.conference.currency.currencyCode = 'USD';
-      $scope.saveEvent();
+    if (
+      !$scope.conference.currency ||
+      !$scope.conference.currency.currencyCode
```
TheNoodleMoose: I'm not sure if this is actually necessary, because if currency is null or undefined then this would be too. But maybe there is a chance currency is defined but the currencyCode didn't get set? I'm not sure, let me know what you think.

PR #683, opened 2020-03-06
EVENT-660 - Blurb for International Currency Selection

caleballdrin:
Added Disclaimer "?" blurb for International Currency Selection


wrandall22:
So, you should look into yarn, and how it works with lint. That should get the Travis build running correctly. Also, you can use yarn to run the project locally to test your change. You can take a screenshot and post it here in the comments.

TheNoodleMoose:
@caleballdrin Yeah to jump onto what @wrandall22 said, the build is failing because we use something called prettier to format the code. If you are using VSCode, you can install the prettier plugin and it will format on save for you(you have to change the settings to do this). If not once you get the project running locally, you can run `yarn prettier:write` and it will format the whole project and fix the error.

wrandall22:
Another thing that would be nice (for convenience and convention) would be to put a link to the Jira ticket in the PR description.

caleballdrin:
I want to merge into staging not master.

PR #687, opened 2020-03-11
EVENT-658: Registration End Date - Allow time zone to be specified

katefree:


app/scripts/services/ConfCache.js
```diff
@@ -77,6 +77,7 @@ angular
           .add(20, 'days')
           .format('YYYY-MM-DD HH:mm:ss'),
         eventTimezone: 'America/New_York',
+        registrationTimezone: 'Asia/Bangkok',
```
OzzieOrca: Why is `'Asia/Bangkok'` the default for new conferences?

katefree: Fixed

PR #691, opened 2020-03-13
Add-Lint-To-Travis-Build

TheNoodleMoose:
Fix lint issues and add lint to travis build. I made this as a suggestion by @wrandall22 .

### DIFF
--- a/test/spec/controllers/eventRegistrations.spec.js
+++ b/test/spec/controllers/eventRegistrations.spec.js
@@ -68,7 +68,7 @@ describe('Controller: eventRegistrations', function() {
     scope.editRegistrant(registrant);
     $httpBackend.flush();
 
-    expect(openModal).toHaveBeenCalled();
+    expect(openModal).toHaveBeenCalled(); /* eslint-disable-line */
 
     $httpBackend.verifyNoOutstandingExpectation();
     $httpBackend.verifyNoOutstandingRequest();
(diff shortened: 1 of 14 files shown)


test/spec/directives/registrationTypeSelectVisibleTypes.spec.js
```diff
@@ -53,19 +53,6 @@ describe('Directive: registrationTypeSelect visibleRegistrantTypes', function()
     expect(typeNames).toContain('Group 1 Dependant 2');
   });
 
-  it('when Group 1 selected on the first screen, only associated registrant types visible on review screen', function() {
```
wrandall22: What was the reason to remove this test?

TheNoodleMoose:
@OzzieOrca What's weird is the travis build shows a bunch of warnings but passes no problem. But locally I run it and get no warnings and it passes fine as well.

test/spec/directives/registrationTypeSelectVisibleTypes.spec.js
TheNoodleMoose: Duplicate test, the one right above it is the same one.

wrandall22:
When I run locally, I get 62 warnings.

TheNoodleMoose:
> When I run locally, I get 62 warnings.

Figured out, ran yarn and now I'm getting the warnings. I'll go ahead and fix these.

test/spec/controllers/eventRegistrations.spec.js
```diff
@@ -68,7 +68,7 @@ describe('Controller: eventRegistrations', function() {
     scope.editRegistrant(registrant);
     $httpBackend.flush();
 
-    expect(openModal).toHaveBeenCalled();
+    expect(openModal).toHaveBeenCalled(); /* eslint-disable-line */
```
wrandall22: Why do we need to disable lint on these lines?

TheNoodleMoose: So it wants us to use `.toHaveBeenCalledWith()` instead of `toHaveBeenCalled()` but the objects these expect are a giant mess and not worth the effort I think to add. As long as it's being called it should be okay.

OzzieOrca:
I think there was some plugin to run eslint as part of the webpack build. But that may have never run on test which it looks like you discovered. We could remove that config or just leave it as I think it show warnings in development when running the dev server.

PR #694, opened 2020-04-03
EVENT-677: Pre-fill staff account number : Web

katefree:


app/scripts/directives/payment.js
```diff
@@ -310,6 +316,19 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function() {
         },
         true,
       );
+
+      function trimEmployeeId() {
```
hlbraddock, line 320: I would suggest renaming `trimEmployeeId` to something more meaningfully indicating why it's doing so, as in something like `convertEmployeeIdToAccountNumber` (where account number in this case is more synonymous with a designation number). @katefree

PR #700, opened 2020-07-15
Okta Login

hlbraddock:
See https://jira.cru.org/browse/EVENT-698
Companion API https://github.com/CruGlobal/ert-api/pull/831


OzzieOrca:
Does a rebranding from The Key to Okta need to happen at this point? Idk what the plan is there.

hlbraddock:
Branding is not yet done, but will be configured server side. I'll likely need to change the login URL at that point.

OzzieOrca:
Guess I meant are we going to keep showing The Key logo on the login button in the modal.

hlbraddock:
I had thought to leave the Relay login brand in the modal as is, since I'm hoping to retain all branding without change, including Okta login page.

Though, I could update login modal branding as per https://signon.cru.org/cas/login, which is what I hope to brand Okta login with.

However, since that https://signon.cru.org/cas/login brand is not currently shown in the login modal, I didn't see the need to change it.

hlbraddock:
@OzzieOrca 
I've updated with new commits since your last review as well as squashing some into previous commits.
The changes introduced are a result of no longer planning to provide Cru/Relay branded login page within Okta, but rather relying on the default generic (possibly with ERT logo) Okta login page (http://signon.okta.com). 
So, these new commits incorporate Okta into the ERT **presentation** layer.

app/scripts/services/logout.js
```diff
@@ -10,10 +10,10 @@ angular
           }
           break;
         case 'RELAY': {
-          // if relay, then redirect to the Relay logout URL
-          const serviceUrl = $location.absUrl().replace('logout', '');
-          $window.location.href =
-            'https://signon.cru.org/cas/logout?service=' + serviceUrl;
+          // if relay, then redirect to the Okta logout URL
+          if (data) {
+            $window.location.href = data.oktaUrl;
```
OzzieOrca, line 15: Does this know how to come back to the ERT or leave you stuck on the Okta site?

hlbraddock: It knows how to come back to the ERT.

hlbraddock:
@OzzieOrca Now available for QA at https://stage.eventregistrationtool.com.

PR #702, opened 2020-07-28
Remove Instragram Login

hlbraddock:
Remove Instagram Login https://jira.cru.org/browse/EVENT-707

Instagram was to be retired as per “The final Instagram Legacy API permission (“Basic”) will be disabled on March 31, 2020"

See https://jira.cru.org/browse/EVENT-684


hlbraddock:
Releasing, with @OzzieOrca approval, despite codecov/path failure, which resulted from lines deleted from files which previously were not sufficiently covered.

PR #725, opened 2020-11-10
EVENT-711: Disable payment type checkboxes and show warning when paym…

OzzieOrca:
…ent details missing

Screen Shot 2020-11-10 at 3 48 40 PM
Screen Shot 2020-11-10 at 3 48 51 PM

Right now those warning triangles always show up and the inputs are disabled on page load if the conditions for them are met. I could make them show up instead when checking an input but that would put it into an invalid state and I couldn't disable the inputs. Might be better to not bother users with them when they aren't using the payment options. I can work on that if you think that's the better strategy.

I was going to say you can play with it at the Netlify link but auth seems to redirect to staging after instead of respecting the clientUrl. There was a domain change from `netlify.com` to `netlify.app` that Netlify changed on us this year. We should probably just move everything to Amplify at some point.

### DIFF
--- a/app/styles/common.scss
+++ b/app/styles/common.scss
@@ -6,21 +6,31 @@
 /*** Buttons ***/
 .btn-primary {
   background: $blue;
-  border-color: #2F84CD;
+  border-color: #2f84cd;
 }
-.btn-primary:hover, .btn-primary:focus, .btn-primary.focus, .btn-primary:active, .btn-primary.active, .open>.dropdown-toggle.btn-primary{
+.btn-primary:hover,
+.btn-primary:focus,
+.btn-primary.focus,
+.btn-primary:active,
+.btn-primary.active,
+.open > .dropdown-toggle.btn-primary {
   background-color: #337ab7;
-  border-color: #2969A0;
+  border-color: #2969a0;
 }
-.btn-info{
+.btn-info {
   background-color: $lightblue;
-  border-color: #318FC2;
+  border-color: #318fc2;
 }
-.btn-info:hover, .btn-info:focus, .btn-info.focus, .btn-info:active, .btn-info.active, .open>.dropdown-toggle.btn-info{
-  background-color: #4FC0EC;
-  border-color: #2BABD0;
+.btn-info:hover,
+.btn-info:focus,
+.btn-info.focus,
+.btn-info:active,
+.btn-info.active,
+.open > .dropdown-toggle.btn-info {
+  background-color: #4fc0ec;
+  border-color: #2babd0;
 }
-.btn-bold{
+.btn-bold {
   @extend .sans-serif-semi;
 }
 .btn-important {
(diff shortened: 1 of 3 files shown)


app/styles/common.scss
```diff
@@ -49,24 +64,33 @@ a.disabled {
...
 
+.warning-popover {
+  cursor: pointer;
+  margin-left: 8px;
+  color: $brand-warning;
+  &:hover {
+    color: $brand-warning;
+  }
+}
+
```
OzzieOrca, line 84: Sorry for all the other changes in this file. Prettier formatted it on commit. These are the only real changes in this file.

app/views/eventDetails/regTypes.html
```diff
@@ -230,40 +230,70 @@ <h4 translate>Group Registration</h4>
...
                   <translate>Accept Account Transfers</translate>
                 </label>
               </div>
               <div class="checkbox">
                 <label>
-                  <input type="checkbox" ng-model="type.acceptScholarships" />
+                  <input
+                    type="checkbox"
+                    ng-model="type.acceptScholarships"
+                    ng-disabled="!type.acceptScholarships && conference.cruEvent && !(conference.businessUnit && conference.operatingUnit && conference.department && conference.projectId)"
```
TheNoodleMoose, line 285: Do you think since these same four variables are used over and over again it would be better to define them as a single variable instead? Or maybe even a function?
```
$scope.paymentTypeIsDisabled = paymentType => {
      return (
        !paymentType &&
        conference.cruEvent &&
        !(
          conference.businessUnit &&
          conference.operatingUnit &&
          conference.department &&
          conference.projectId
        )
      );
    };
```

OzzieOrca: AngularJS makes this too hard. But ya probably. I can do that.

hlbraddock: ReactJS rewrite being scheduled sometime in the mid term future.

hlbraddock: Note that not all of these chart field components need to be set in order for it to be considered a valid chart field. For some chart fields, it's possible for `projectId` to be blank. I don't know if there may be other chart fields where some other components might also be empty.

Ultimately, the ert-api will validate on conference save, so the save request will be denied if the chart fields are not valid and/or if the conference is including support for payments without any chart field set.

You can still attempt front end validation, if you deem that a better user experience (you'll still get some semi meaningful error message as response from ert-api, namely `Conference accepting any type of payment must provide chart field`) but at the very least, it cannot include requiring `projectId` be set.

hlbraddock:
Note that this work is currently in QA. And final formal QA signoff is scheduled for 12/10. 

Please finish PR modifications prior to that time.

Thanks! @OzzieOrca

app/views/eventDetails/regTypes.html
OzzieOrca: Looks like I updated one to consider it valid when any field was provided but I forgot to update the others. Guess that's another argument to write a function haha. Thanks for catching that.

TheNoodleMoose (approved):

app/views/eventDetails/regTypes.html
hlbraddock: I've been informed that only Project ID would be _possible_ (some chart fields require it, some don't) to be left blank in a valid chart field.

hlbraddock: However, having said that, I'm now thinking that it should not be incumbent on the UI to understand valid PS Financials chart field validation rules. The ert-api doesn't know directly either, but must enquire of PS Financials in order to determine. 

Thoughts?

OzzieOrca: Seems like we could either:
1. Require `businessUnit`, `operatingUnit`, and `department` to match PS Financials
1. Require `businessUnit`, `operatingUnit`, or `department` to just make sure they entered something

I'm assuming it'd be unlikely for these rules to change?

hlbraddock: ert-api calls out to PS Financials endpoint to validate. 
Option 2 seems sufficient.
I don't think the rules will change.

OzzieOrca:
I merged this to staging.

PR #726, opened 2020-11-17
EVENT-726: Add Google Tag Manager (GTM)

OzzieOrca:
[Jira](https://jira.cru.org/browse/EVENT-726)

### DIFF
--- a/app/index.ejs
+++ b/app/index.ejs
@@ -47,11 +47,7 @@
           j = d.createElement(s),
           dl = l != 'dataLayer' ? '&l=' + l : '';
         j.async = true;
-        j.src =
-          'https://www.googletagmanager.com/gtm.js?id=' +
-          i +
-          dl +
-          '&gtm_auth=VMsIb-u97IZIyagKFgXe1w&gtm_preview=env-1&gtm_cookies_win=x';
+        j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl;
         f.parentNode.insertBefore(j, f);
       })(window, document, 'script', 'dataLayer', 'GTM-5VHJ5WH');
     </script>
@@ -129,7 +125,7 @@
     <!-- Google Tag Manager (noscript) -->
     <noscript
       ><iframe
-        src="https://www.googletagmanager.com/ns.html?id=GTM-5VHJ5WH&gtm_auth=VMsIb-u97IZIyagKFgXe1w&gtm_preview=env-1&gtm_cookies_win=x"
+        src="https://www.googletagmanager.com/ns.html?id=GTM-5VHJ5WH"
         height="0"
         width="0"
         style="display:none;visibility:hidden"
--- a/app/index.ejs
+++ b/app/index.ejs
@@ -22,7 +22,6 @@
       </script>
     <![endif]-->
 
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/reviewRegistration.js
```diff
@@ -164,10 +164,10 @@ angular
           return registration.completeRegistration(currentRegistration);
         })
         .then(function() {
-          //send analytics event
-          analytics.digitalData.eventID = conference.id;
-          analytics.digitalData.registeredEventName = conference.name;
-          analytics.track('registration');
+          analytics.track('registration', {
```
TheNoodleMoose, line 167: Although I think this is a better way to do it, I think Clark just wants us to send the event name ```registration``` and then access the digital Data layer from the GTM side like we were doing or Adobe.

app/index.ejs
```diff
@@ -36,10 +35,49 @@
...
+        w[l].push({ 'gtm.start': new Date().getTime(), event: 'gtm.js' });
+        var f = d.getElementsByTagName(s)[0],
+          j = d.createElement(s),
+          dl = l != 'dataLayer' ? '&l=' + l : '';
+        j.async = true;
+        j.src =
+          'https://www.googletagmanager.com/gtm.js?id=' +
+          i +
+          dl +
+          '&gtm_auth=VMsIb-u97IZIyagKFgXe1w&gtm_preview=env-1&gtm_cookies_win=x';
```
wrandall22, line 54: I don't think we want `gtm_preview` in prod?

TheNoodleMoose:
@OzzieOrca Talking with Brenton and Clark, they decided to not go with the history change to trigger page views, and instead use a custom event(virutal-page-view) to trigger one. We just need to add this event to the ```firePageViewEvent``` function, with the event name being ```virtual-page-view```.
Screen Shot 2020-11-18 at 9 31 08 AM

app/index.ejs
OzzieOrca: I got that code from the Live environment at https://tagmanager.google.com/?authuser=1#/admin/accounts/1597949959/containers/35270741/environments?containerDraftId=2 but I'll use the one from the link on the main dashboard page.

app/scripts/controllers/reviewRegistration.js
OzzieOrca: Would setting both ways hurt? Seems like it'd be better not to mutate global state.

TheNoodleMoose: Technically you can send the data and they can just ignore it on the gtm side, but still access the digitalData layer like before.

OzzieOrca:
Was there a problem with history triggers or are they just used to the other way? Seems like less custom code in our apps would be better.

TheNoodleMoose:
They just want to use it the other way since they know it will work. I guess Sarah Rose(our consultant from adswerve) said that the history change can be hit or miss so they don't want to give it a try. I agree, the less custom code the better.

TheNoodleMoose:
Brenton mentioned maybe trying the history change after the migration is completed when we are cleaning up the adobe code.

TheNoodleMoose (approved):

PR #727, opened 2020-12-02
add-additional-meta-properties

TheNoodleMoose:
I noticed there was some warnings in the admin menu on Facebook developers saying we should add these. So I went ahead and did so.
Screen Shot 2020-12-02 at 3 15 03 PM

### DIFF
--- a/app/index.ejs
+++ b/app/index.ejs
@@ -8,6 +8,27 @@
       name="description"
       content="The Event Registration Tool provides everything you need when it comes to event registration and management."
     />
+    <meta
+      property="og:url"
+      content="https://www.eventregistrationtool.com/"
+    />
+    <meta property="og:type" content="website" />
+    <meta
+      property="og:title"
+      content="Event Registration Tool"
+    />
+    <meta
+      property="og:image"
+      content="https://eventregistrationtool.com/assets/favicon.ico"
+    />
+    <meta
+      name="og:description"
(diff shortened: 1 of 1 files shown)


app/index.ejs
```diff
@@ -8,6 +8,27 @@
...
+      property="og:title"
+      content="Event Registration Tool"
+    />
+    <meta
+      property="og:image"
+      content="https://eventregistrationtool.com/assets/favicon.ico"
+    />
+    <meta
+      name="og:description"
+      content="The Event Registration Tool provides everything you need when it comes to event registration and management."
```
OzzieOrca, line 26: I wonder if this should be a more registrant friendly message. I assume the most common way snippets would be shown is when people share registration links. This doesn't quite sound like you're clicking the link in order to register.

OzzieOrca (approved):

app/index.ejs
TheNoodleMoose: I came up with this a while back when I saw the javascript code is set as the description when it is displayed as a search result on google. I asked Mike Schmitt for a better description a while ago but he never got back to me.

TheNoodleMoose: @OzzieOrca Do you think this is good as is? We could always update it once Mike Schmitt gets back to me with an updated version of the description or if you have any suggestions on better verbiage.

OzzieOrca: It's not a regression so feel free to merge it :)

PR #736, opened 2021-04-14
EVENT-736-Journal-Upload-Page

TheNoodleMoose:
Jira ticket: https://jira.cru.org/browse/EVENT-736

@OzzieOrca I've been working on this for a while now so it's probably a good time to get some fresh eyes on it. Let me know if you have any questions about the logic or if you see anything that just looks off/could be improved.


TheNoodleMoose:
Also, note that this logic is currently being QA'd and additional changes could arise. But I figured it would be good to have someone start to take a look sooner than later.

app/scripts/controllers/journalUpload.js
```diff
@@ -0,0 +1,241 @@
...
+      if (newReportId !== oldReportId) {
+        $window.scrollTo(0, 0);
+      }
+      // if currentReportId is null, refresh account transfer data
+      if (!newReportId) {
+        $scope.refresh();
+      } else {
+        // set accountTransfers to current report's valid account transfers
+        $scope.accountTransfers = $scope.reports
+          .filter(report => report.id === newReportId)[0]
```
OzzieOrca, line 96: Can this find multiple?  Seems like `.find` would be better.
```suggestion
          .filter(report => report.id === newReportId)[0]
```

Also maybe part of the code could be shared with the error one below.

app/scripts/controllers/journalUpload.js
```diff
@@ -0,0 +1,241 @@
...
+
+    $scope.allAccountTransfersIncluded = () =>
+      $scope.accountTransfersToInclude.length ===
+      $scope.accountTransfers.length;
+
+    $scope.addAllTransfersToInclude = () => {
+      angular.forEach($scope.accountTransfers, accountTransfer => {
+        if ($scope.accountTransfersToInclude.indexOf(accountTransfer) === -1) {
+          accountTransfer.included = true;
+          $scope.accountTransfersToInclude.push(accountTransfer);
```
OzzieOrca, line 153: Little odd that you're maintaining this state in two places.

app/scripts/services/journalUploadService.js
```diff
@@ -0,0 +1,139 @@
...
+        filterPayment: '',
+        filterRegType: '',
+        includeAccountTransfers: true,
+        includeCheckedin: 'yes',
+        includeWithdrawn: 'yes',
+        includeIncomplete: 'yes',
+        primaryRegistrantOnly: true,
+      },
+    ) => {
+      const defer = $q.defer();
```
OzzieOrca, line 33: You shouldn't need to use `defer`, just return the promise and return values/rethrow errors in the promise then/catch.

Applies to other lines too.

app/scripts/services/journalUploadService.js
```diff
@@ -0,0 +1,139 @@
...
+
+      return defer.promise;
+    };
+
+    this.getAccountTransferData = data => {
+      return _.flatten(
+        data.registrations.reduce((result, registrant) => {
+          return [
+            ...result,
+            registrant.accountTransfers.filter(
```
OzzieOrca, line 93: Just spread this, then you don't have to call `flatten`:
```suggestion
            ...registrant.accountTransfers.filter(
```

Applies to other lines too.

app/scripts/services/journalUploadService.js
```diff
@@ -0,0 +1,139 @@
...
+      $rootScope.loadingMsg = 'Submitting account transfers';
+      // Filter out included key from account transfer object
+      const filteredAccountTransfers = accountTransfers.map(accountTransfer => {
+        return Object.keys(accountTransfer)
+          .filter(key => key !== 'included')
+          .reduce((obj, key) => {
+            obj[key] = accountTransfer[key];
+            return obj;
+          }, {});
+      });
```
OzzieOrca, line 124:
```suggestion
      const filteredAccountTransfers = accountTransfers.map(({ included, ...accountTransfer }) => accountTransfer);
```

OzzieOrca (changes requested):
That's the most new Angular code I've seen in a while haha 🙂  Hopefully it wasn't too bad...

app/scripts/services/journalUploadService.js
TheNoodleMoose: I had to add the babel plugin ```transform-object-rest-spread``` and then update the eslint rule: 
```
"no-unused-vars": [
      "error",
      { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }
    ],
```

but it works now 🙂

app/scripts/controllers/journalUpload.js
```diff
@@ -0,0 +1,241 @@
...
+    $scope.authToken = $cookies.get('crsToken');
+
+    $scope.$watch(
+      'queryParameters',
+      (newParameters, oldParameters) => {
+        //limit changed, reset page
+        if (
+          newParameters.page > 1 &&
+          newParameters.page === oldParameters.page
+        ) {
```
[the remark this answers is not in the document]
TheNoodleMoose: I based this on how it works for the existing payments page: https://github.com/CruGlobal/conf-registration-web/blob/db6dd278a4d632890f15f2bd00c6556874e6020b/app/scripts/controllers/paymentReport.js#L41-L44
Technically if they change the limit, that will cause a refresh of the data.  But I think what's of most concern is resetting them back to the first page if they currently are not already on it which is why ```page``` was chosen in favor ```limit``` but I think it could work either way really.

app/scripts/services/journalUploadService.js
TheNoodleMoose: Gotcha, it's interesting that there's so many different ways it's done in this codebase haha

app/scripts/controllers/journalUpload.js
TheNoodleMoose: It is a bit, once they uncheck to include the account transfer it's removed from ```accountTransfersToInclude``` and the ```included``` value for that account transfer is set back to ```false```. So really the included state only matters in ```$scope.accountTransfers```. Not too sure how to improve it, as it's needed for the checked/unchecked state of the checkbox element.

app/scripts/services/journalUploadService.js
TheNoodleMoose: But it breaks Karma 😞

app/scripts/controllers/journalUpload.js
TheNoodleMoose: I changed it to ```.find``` and was like "Why didn't I do this the first time?" and then realized when I originally used it, it breaks all the tests 🙁 I could use the lodash ```_.find``` but I'm not sure if that's any better.

OzzieOrca (approved):
Looks good. Ya there's so much historic stuff in this repo. If you want a look at a better structured AngularJS app, you could look at give-web or mpdx_web.

That's annoying karma doesn't support the newer stuff. Sorry for the wild goose chase there. Maybe there's config for it. A Jest migration would be nice someday. But then again a complete rewrite would be good too.

OzzieOrca (changes requested):

app/scripts/controllers/paymentModal.js
```diff
@@ -449,24 +460,9 @@ angular
...
-              pastPayment.paymentType === 'TRANSFER') &&
-            pastPayment.reported === true,
-        );
-        return filteredPastPayments.length > 0;
-      }
-    };
+    // Disable editing all other fields if payment has been reported and the paymentType is either SCHOLARSHIP or TRANSFER
+    $scope.disableEditingFieldsForReportedPayments = payment =>
+      payment.reported &&
+      (payment.paymentType === 'SCHOLARSHIP' ||
```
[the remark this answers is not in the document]
TheNoodleMoose: Actually, I need to disable all other fields if the payment has been reported and IT IS either a scholarship or transfer.

app/scripts/controllers/paymentModal.js
```diff
@@ -146,6 +146,17 @@ angular
...
+    $scope.canEditPayment = payment => {
+      if (payment.reported) {
+        return (
+          payment.paymentType === 'SCHOLARSHIP' ||
+          payment.paymentType === 'TRANSFER'
+        );
+      } else {
+        return true;
+      }
+    };
```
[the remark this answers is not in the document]
TheNoodleMoose: It's similar to a degree. ```canEditPayment``` says if it's not reported regardless of paymentType allow them to edit, or if it has been reported, only allow editing if the paymentType is ```SCHOLARSHIP``` or ```TRANSFER```. ```disableEditingFieldsForReportedPayments``` says the payment has to be reported and it has to be either a ```SCHOLARSHIP``` or a ```TRANSFER```.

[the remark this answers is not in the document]
OzzieOrca: You sure these shouldn't be inverses of each other?

What I'm reading with `canEditPayment` is that the fields are **ENABLED** if any of these conditions:
1. unreported
2. SCHOLARSHIP
3. TRANSFER

What I'm reading with `disableEditingFieldsForReportedPayments` is that the fields are **ENABLED** if any of these conditions:
1. unreported
2. not SCHOLARSHIP and not TRANSFER (basically you can edit any payment type that isn't SCHOLARSHIP or TRANSFER)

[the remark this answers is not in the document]
TheNoodleMoose: All ```canEditPayment``` does is hide the edit button in the UI. ```disableEditingFieldsForReportedPayments``` disables all other fields if they meet the conditions:

OzzieOrca (approved):

app/scripts/controllers/journalUpload.js
```diff
@@ -76,8 +77,10 @@ angular
           //scroll to top on page change
           $window.scrollTo(0, 0);
         }
-
-        $scope.refresh();
+        // Only refresh when paramters are different
+        if (newParameters !== oldParameters) {
+          $scope.refresh();
```
TheNoodleMoose: This was always just refreshing no matter what, so on page load, there was an additional API call for account transfers.

app/scripts/controllers/journalUpload.js
```diff
@@ -87,6 +90,12 @@ angular
       if (newReportId !== oldReportId) {
         $window.scrollTo(0, 0);
       }
+
+      // Prevent refresh on initial page load
+      if (!newReportId && !oldReportId) {
+        return;
+      }
```
TheNoodleMoose, line 97: A similar thing here on page load is the only time that the ```oldReportId``` and ```newReportId``` should both be null. This will prevent an additional API call.

OzzieOrca (approved):

app/views/journalUpload.html
```diff
@@ -84,7 +84,6 @@ <h2 class="page-title" translate>
           <option value="full" translate>Full/Overpaid</option>
           <option value="partial" translate>Partial</option>
           <option value="full-partial" translate>Full/Partial</option>
-          <option value="none" translate>Not Paid</option>
```
TheNoodleMoose: I was asked to remove this as it would be confusing to leave in. Anything they should see on the Journal Upload page should technically be "Paid".

app/views/journalUpload.html
```diff
@@ -288,9 +288,6 @@ <h4>
             <th>
               <a href="" translate>GL Account</a>
             </th>
-            <th>
-              <a href="" translate>Product Code</a>
```
TheNoodleMoose: Another QA request. I was asked to remove this as some backend correction could happen in people soft that would not be reflected in the ERT. So instead of possibly showing an incorrect product code, we'll just remove the column.

OzzieOrca (approved):

app/views/eventDetails/paymentOptions.html
```diff
@@ -228,6 +228,14 @@ <h3 class="panel-title" translate>Cru Event Chartfield</h3>
 
       <div class="col-sm-3">
         <label translate>Project Id</label>
+        <i
+          class="fa fa-question-circle"
+          button
+          uib-popover="{{'If there is no Project Id (or `NONE`), leave blank' | translate}}"
```
TheNoodleMoose, line 234: I wanted to do ```'If there is no Project Id (or "NONE"), leave blank'```, but the extra set of double quotes around NONE causes an error in the template. Does the same thing for extra single quotes 🙁 I figured having something around it would be better than nothing. There's also the option of defining as a variable ```$scope.projectIdHoverText = 'If there is no Project Id (or "NONE"), leave blank' ``` and then doing ``` {{ projectIdHoverText | translate }}``` but wasn't sure how I felt about that. Let me know what you think.

OzzieOrca: Can you escape the quotes? `\"`

https://stackoverflow.com/questions/22967475/angularjs-html-in-js-escaping-single-quote

I'm not sure what works without trying it.

app/scripts/controllers/journalUpload.js
```diff
@@ -55,9 +55,7 @@ angular
       includeIncomplete: 'yes',
       primaryRegistrantOnly: true,
     };
-    $scope.meta = {
-      totalPages: 0,
-    };
+    $scope.meta = registrationsData.meta;
```
TheNoodleMoose, line 58: So there was a bug that happened after I removed the duplicate api calls. Basically, we weren't updating ```$scope.meta``` until the user caused the query parameters to change and we ran ```refresh```. But we already had that metadata, so I just use it here on page load now.

app/scripts/controllers/journalUpload.js
```diff
@@ -11,7 +11,7 @@ angular
     $window,
     $cookies,
     $uibModal,
-    registrations,
+    registrationsData,
```
TheNoodleMoose, line 14: I figured ```registrationsData``` would be a better name instead of having to do ```registrations.registrations```. Let me know what you think.

app/scripts/controllers/journalUpload.js
```diff
@@ -178,7 +177,7 @@ angular
     };
 
     $scope.getRemainingBalance = registrationId => {
-      return _.find($scope.registrations.registrations, {
+      return _.find($scope.registrations, {
```
TheNoodleMoose, line 180: So before ```$scope.registrations``` was the whole response from the api, but now it's specifically just the registration data we care about.

app/scripts/controllers/journalUpload.js
```diff
@@ -130,13 +128,14 @@ angular
     $scope.refresh = () => {
       journalUploadService
         .getRegistrationData(conference.id, $scope.queryParameters)
-        .then(data => {
-          $scope.meta = data.meta;
+        .then(registrationsData => {
+          $scope.meta = registrationsData.meta;
+          $scope.registrations = registrationsData.registrations;
```
TheNoodleMoose, line 133: I wasn't updating the registrations data before, but now I am.

app/views/eventDetails/paymentOptions.html
TheNoodleMoose: That seems to throw a parse error 🙁

OzzieOrca: Idk then. Sorta a trivial thing. Your current solution is probably fine.

OzzieOrca (approved):

OzzieOrca (approved):

OzzieOrca (approved):

PR #750, opened 2021-06-02
update-submitAccountTransfers-error-message

TheNoodleMoose:
Basically, the journal upload submission is taking a long time to process, depending on the number of account transfers included. This is causing a gateway timeout issue. A temporary fix, until API optimization can be done, is to just update the error message for the users to make sure they don't accidentally resubmit.

### DIFF
--- a/app/scripts/services/journalUploadService.js
+++ b/app/scripts/services/journalUploadService.js
@@ -2,6 +2,7 @@ angular
   .module('confRegistrationWebApp')
   .service('journalUploadService', function JournalUploadService(
     $rootScope,
+    $route,
     $http,
     $q,
     modalMessage,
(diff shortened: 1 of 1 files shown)


OzzieOrca (approved):
Looks fine. My only thought would be adding a check for a specific status code but that may not be necessary.

TheNoodleMoose:
@OzzieOrca It actually doesn't return an error code since it times out, which is what made it hard to figure out the issue in the first place.

OzzieOrca:
Oh the browser times out, not the server. Got it. I still think there might be some sort of specific error message/code to match against but maybe that's chrome internals...

TheNoodleMoose:
No, the browser doesn't timeout, the gateway times out but not the ERT API layer. Not sure of the specifics of that, but I do know that a response is not given back.

OzzieOrca (approved):

PR #754, opened 2021-07-30
EVENT-751-DocuSign-Integration

TheNoodleMoose:


webpack.config.js
```diff
@@ -109,23 +109,29 @@ module.exports = (env = {}) => {
...
+          test: /\.(ts|js)x?$/,
           exclude: /node_modules/,
           use: [
             {
               loader: 'babel-loader',
               options: {
                 presets: [
                   [
-                    'env',
+                    '@babel/preset-env',
```
OzzieOrca, line 120: Since you made a `babel.config.js` file, can you move all these options to that file?

Maybe the `isTest` part needs to stay here. Or maybe there's a way to load that from env vars outside the webpack config.

webpack.config.js
```diff
@@ -109,23 +109,29 @@ module.exports = (env = {}) => {
...
                     {
                       modules: false,
-                      exclude: ['transform-es2015-function-name'],
+                      targets: {
+                        browsers: ['last 1 version', 'ie >= 11'],
+                      },
```
OzzieOrca, line 125: I think the defaults are usually fine, this seems like a small target. But sounds like specifying nothing isn't the same as specifying defaults with babel https://babeljs.io/docs/en/babel-preset-env#no-targets.

So this?
```suggestion
                      targets: 'defaults',
```

`defaults` is `> 0.5%, last 2 versions, Firefox ESR, not dead` according to https://github.com/browserslist/browserslist#full-list

app/scripts/controllers/eventDetails.js
```diff
@@ -449,6 +450,18 @@ angular
...
       }
 
+      if (
+        $scope.conference.cruEvent &&
+        $scope.conference.ministry &&
+        $scope.getActivities().length !== 0 &&
+        !$scope.conference.ministryActivity
+      ) {
+        validationErrors.push(
+          'Please enter which Ministry Activitiy is applicable for this event.*',
```
OzzieOrca, line 460: What's the `*`?

app/scripts/main.js
```diff
@@ -4,6 +4,10 @@ import 'scripts/app.config.js';
 import 'scripts/app.run.js';
 import 'scripts/app_enforceAuth.js';
 
+// components
+import 'scripts/components/FormStatusModal/FormStatusModal';
+import 'scripts/components/FormStatusPopover/FormStatusPopover';
```
OzzieOrca, line 9: I was going to suggest importing these where they are used. Although we're not really using the module system right and everything is a global... 

MPDX is making a module for each component https://github.com/CruGlobal/mpdx_web/blob/master/src/home/weeklyActivity/weeklyActivity.component.ts#L70 so at least it would throw an error if dependencies aren't included for each import.

Maybe we should leave it alone for now.

package.json
```diff
@@ -20,24 +19,39 @@
     "moment": "^2.19.3",
     "moment-timezone": "^0.5.27",
     "pickadate": "^3.5.6",
+    "prop-types": "^15.7.2",
```
OzzieOrca, line 22: Do you need prop-types?  Looked like you were using TS.

package.json
```diff
@@ -77,7 +88,8 @@
     "build": "webpack -p",
     "build:analyze": "webpack -p --env.analyze",
     "test": "karma start",
-    "lint": "eslint '@(app|test)/**/*.js'",
+    "lint": "eslint '@(app|test)/**/*.{js,ts,tsx}'",
+    "lint:ts": "tsc",
```
OzzieOrca, line 92: You should make Travis run this. (Or Github Actions hehe)

OzzieOrca (changes requested):
Didn't know you were adding React. Cool! Looks pretty good, just some tooling config suggestions.

package.json
TheNoodleMoose: Ah yeah, I think I added that as I was following the initial setup from https://github.com/CruGlobal/give-web/commits/react that Bill shared with me. I'll remove it.

app/scripts/main.js
TheNoodleMoose: I may address it in a separate pr in the future.

app/scripts/controllers/eventDetails.js
TheNoodleMoose: I think it's to indicated that it's only applicable to certain ministries. All the other ```Ministry``` related validation errors have it so I just copied it from that https://github.com/CruGlobal/conf-registration-web/blob/7dcfbe90bd65bbd8dfee3c1bdf155a76c2ea2d29/app/scripts/controllers/eventDetails.js#L466

package.json
TheNoodleMoose: Ah yep! Github actions eventually 🙂

OzzieOrca (approved):

PR #760, opened 2021-09-29
Migrate-to-GitHub-Actions

TheNoodleMoose:


.github/workflows/sync.yml
```diff
@@ -0,0 +1,52 @@
...
+  sync-master:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 15
+      - name: Install awscli
+        run: pip install --user awscli
```
OzzieOrca, line 18: Looks like the container may already have this installed https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#cli-tools.

You probably don't need to install it. I wonder if it's already added to the path. You might be able to do `aws` instead of `~/.local/bin/aws`.

.github/workflows/sync.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  push:
+    branches: [master, staging]
+  workflow_dispatch:
+
+jobs:
+  sync-master:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
OzzieOrca, line 11: I think you could move this `if` to just the deploy step just to reduce repetition. The rest of the steps are the same between the two.

.github/workflows/sync.yml
```diff
@@ -0,0 +1,52 @@
+name: Sync
```
OzzieOrca, line 1: I would call this Deploy. `sync` is just an implementation detail of the `awscli`. I thought you were referring to the Github PR synchronize event at first.

.github/workflows/sync.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  push:
+    branches: [master, staging]
+  workflow_dispatch:
+
+jobs:
```
OzzieOrca, line 8: I think we want to make sure all our CI checks run successfully before deploying to avoid shipping broken code to production. Since it seems like there's no way to make one workflow depend on another and your `on` events are almost the same, maybe we should move this into the main `ci-cd.yml` file with `needs` for dependencies and some more `if`s. https://github.com/CruGlobal/missionhub-react-native/blob/develop/.github/workflows/ci-cd.yml has some exaples of `needs` depending on other job successes and `if` checking for branches and non PRs (just push events).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
...
+        run: yarn prettier:check
+  yarn-check-cache:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 15
+      - name: 📦 Install Dependencies
+        run: yarn install --immutable --immutable-cache --check-cache
```
OzzieOrca, line 69: I think this is sorta a Yarn v2/Yarn PnP thing since the cache is in the repo instead of being pulled from npm/cached on the CI. But maybe doesn't hurt to leave it. Maybe the repo will get to Yarn v2/v3 someday.

Idk if it really applies to us since I think we mostly trust our collaborators but I just added it since it was a simple job https://yarnpkg.com/features/zero-installs/#does-it-have-security-implications

.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
...
+    runs-on: ubuntu-latest
+    name: tests
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          # Help Codecov detect commit SHA
+          fetch-depth: 2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 15
```
OzzieOrca, line 21: Let's just go all the way to node 16 since this is a new environment. Unless something doesn't work, then it may be easier to just leave it for now.

.github/workflows/sync.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: 📦 Install Dependencies
+        run: yarn install --immutable --immutable-cache
+      - name: 🔨 Build
+        run: yarn build
+      - name: 🚀 Sync to S3
+        run: |
+          ~/.local/bin/aws s3 sync dist s3://cru-ert-web-prod --region us-east-1 --acl public-read --exclude '*.html' --cache-control 'public, max-age=31536000',
+          ~/.local/bin/aws s3 sync dist s3://cru-ert-web-prod --region us-east-1 --acl public-read --include '*.html' --cache-control 'public, max-age=180'
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
OzzieOrca, line 28: Looks like you don't have these setup at https://github.com/CruGlobal/conf-registration-web/settings/secrets/actions yet. Might be a good excuse to rotate the keys since Travis won't need them after this is merged.

The #aws Slack channel has an email in the channel description you can use for getting help creating keys (thought dumping the email in a public repo might be weird). They can delete the old keys after we confirm the Github Actions deploy is working.

OzzieOrca (changes requested):
Looks pretty good, thanks for setting this up :) I left a few comments. Let me know if you have any questions :)

.github/workflows/sync.yml
TheNoodleMoose: Yeah, I was thinking it would be good to wait to get them until this pr is good to go. But thanks for letting me know how to go about getting new ones because I was unsure 🙂

.github/workflows/sync.yml
TheNoodleMoose: Yeah, when I was researching I found two different examples, one where they installed and the other they didn't. I went with this first because it was closer to what we currently had. But I'm all for simplifying where we can so I'll remove the install step and try using just ```aws```.

.github/workflows/ci.yml
TheNoodleMoose: Yeah 16 breaks things haha

OzzieOrca: `node-sass` seems to need an upgrade for every node version 😢 It's been a pain.

Thankfully migrating to dart-sass (npm `sass` package) no longer needs a native build step.
https://github.com/sass/node-sass#node-sass
> Warning: LibSass and Node Sass are deprecated. While they will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto Dart Sass.

I've migrated a couple repos:
https://github.com/CruGlobal/mpdx_web/pull/2232 https://github.com/CruGlobal/mpdx_web/pull/2235
https://github.com/CruGlobal/campus-contacts-web/pull/837

Doesn't need to happen in this PR but that's how to do it :) Hopefully we don't have to fix too many things but I have no idea what the state of the sass is in this project. It should spit out warnings for everything.

I used
```js
                    {
                        loader: 'sass-loader',
                        options: {
                            sassOptions: {
                                quietDeps: true,
                            },
                        },
                    },
```
to swallow warnings in NPM packages that had outdated sass files.

OzzieOrca (approved):
Looks great :)

.github/workflows/ci.yml
```diff
@@ -85,8 +85,9 @@ jobs:
         env:
           ROLLBAR_ACCESS_TOKEN: ${{ secrets.ROLLBAR_ACCESS_TOKEN }}
       - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v2
+        uses: aws-actions/configure-aws-credentials@v1
```
OzzieOrca, line 88: I was about to ask why v2 didn't work but it appears v1 is the latest/only published version 😆

OzzieOrca:
Idk how/when you intend on testing this but we should try a merge into the staging branch first before merging into master. Also we need to update the required Github status checks eventually. Let me know if you don't have permission to do that.

PR #765, opened 2021-10-18
EVENT-762-Integrate-Google-Signin

TheNoodleMoose:
This PR adds the ability to sign in with Google! I also went ahead and updated the other sign-in method buttons to provide a consistent and uniform look and feel. Looks pretty good if I do say so myself 😉
Screen Shot 2021-10-18 at 3 23 02 PM


app/views/modals/loginDialog.html
```diff
@@ -1,4 +1,5 @@
-<div class="modal-header">
+<!-- <script src="https://accounts.google.com/gsi/client" async defer></script> -->
```
OzzieOrca, line 1: This still needed?

OzzieOrca (approved):

PR #772, opened 2022-01-11
Migrate-to-dart-sass

TheNoodleMoose:
Migrating in order to fix issue @hlbraddock was running into with  getting ERT to install/run locally on his new mac https://github.com/sass/node-sass/issues/3033


OzzieOrca (approved):
Does sass have any warnings? I assume we have a lot of really old scss files but maybe we weren't using very many complex features.

TheNoodleMoose:
@OzzieOrca I didn't see any. I went through the scss files to see if there was any of the math logic that you had to make changes to but didn't find anything.

PR #776, opened 2022-01-19
EVENT-785-First-Page-Visible-When-No-Content-Applies

TheNoodleMoose:
Ticket: https://jira.cru.org/browse/EVENT-785

Basically, two bugs were happening and this fix resolves them both. If the first page is empty, as in no questions, when the user started registering they would land on that first page and it would be empty. The second has a similar end result of an empty page, but was resulted because there were no questions on the first page for the specific ```registrantType```. The ticket shows this in more detail.

So my fix was to filter out all pages that have no questions blocks. This was happening in some capacity already, but the links navigated to the first page, regardless of if it had any valid question blocks or if it was ```visible```.

### DIFF
--- a/app/scripts/controllers/registration.js
+++ b/app/scripts/controllers/registration.js
@@ -41,13 +41,26 @@ angular
     $scope.activePageIndex = _.findIndex($scope.conference.registrationPages, {
       id: pageId,
     });
+    // Filter through all pages and remove any empty pages
+    // Check for registrant needed since on the welcome page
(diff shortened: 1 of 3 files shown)


OzzieOrca (approved):
Seems like the API could filter these out or the question editor could prevent empty pages but I'm ok with this. There's a lot of ERT business logic pushed client-side.

TheNoodleMoose:
Yeah, that may be an improvement I visit during the rewrite.

PR #779, opened 2022-01-27
EVENT-776-Display-Foreign-Addresses

TheNoodleMoose:
This PR adds a new country input field for the event address and address question. For simplicity's sake, we made it to where if the country is non-US, the state select input just becomes a text input and the placeholders change to 'State/Region' and 'ZIP/Postal Code'. We do have to account that not every foreign address will have a state/region or even a zip, so we provide verbiage via a tooltip to let the user know they only need to fill in the applicable fields. Also if an address question is required, we don't want to require the state or zip part just in case it does not apply to them.


app/scripts/controllers/eventDetails.js
```diff
@@ -81,6 +81,9 @@ angular
     $scope.originalConference = conference;
     $scope.conference = angular.copy(conference);
     $scope.currencies = currencies;
+    $scope.conference.locationCountry = conference.locationCountry
+      ? conference.locationCountry
+      : 'US';
```
TheNoodleMoose, line 86: I think the API will set this to ```''```(@Benjamin-deVries can confirm) by default, so it's up to the UI to set the value to 'US'.

app/scripts/controllers/eventDetails.js
```diff
@@ -690,6 +693,17 @@ angular
       }
     });
 
+    $scope.$watch(
+      'conference.locationCountry',
```
TheNoodleMoose, line 697: If the country changes, we should just wipe the values for state and zip so avoid any possible confusion and allow the user to input was they need for their specific case.

wrandall22:
One initial thought I have is that Siebel and give site use address line 1, 2, 3, 4 for international addresses. Not sure if we want to be consistent, maybe it's not worth it.

app/scripts/directives/blockRegistration.js
```diff
@@ -87,6 +87,14 @@ angular
...
                 return;
               }
+              // clear saved value of answer state and zip if address question and country changes
+              if (
+                answer.value.country &&
+                answer.value.country !== oldAnswer.value.country
+              ) {
+                answer.value.state = '';
+                answer.value.zip = '';
+              }
```
TheNoodleMoose, line 97: I really don't like having to do this at all. The current logic didn't do any checks like this, so having to add this for a specific use case feels meh. 

The reason behind needing this is, say I change the country on my address question to 'UZ' from 'US', and then fill in the state input with ```Cool State```. If I change the country back to 'US', the input switches back to a select and it can't find a valid option for ```Cool State``` so it renders blank. But it has a value so from the underlying logic, the form is valid and the user can continue with an invalid US state. Maybe this is an extra edge case, but I think it is one that could occur and needs to be covered. Just wish it could happen some other way. Opinions or suggestions for a different approach are welcomed 🙂

TheNoodleMoose:
@wrandall22 So essentially ```city```, ```state```, and ```zip``` become generic address fields(```Address Line 1```, ```Address Line 2```, etc) that are text inputs? I do kind of like that idea. @hlbraddock do you have opinions on this? I could also bring this idea up to Charles and Mike to see what they think.

app/scripts/directives/blockRegistration.js
TheNoodleMoose: One thought I just had would be to have a check when submitting the form that if the address question is required and the country is US, then the state value has to be one of the 50 states. Could just have this be an array and check that the value of state is included in that array.

wrandall22:
City, State, Zip becomes address lines 3 and 4. Totally different fields tracked on the backend.

hlbraddock:
> @wrandall22 So essentially `city`, `state`, and `zip` become generic address fields(`Address Line 1`, `Address Line 2`, etc) that are text inputs? I do kind of like that idea. @hlbraddock do you have opinions on this? I could also bring this idea up to Charles and Mike to see what they think.

We captured initial address requirements in https://jira.cru.org/browse/EVENT-776. If you have other ideas, we may want to confer with Product/Stakeholder once more.

app/scripts/services/constants.js
```diff
@@ -10,3 +10,57 @@ angular.module('confRegistrationWebApp').constant('ruleTypeConstants', {
   FORCE_SELECTION: 'FORCE_SELECTION',
   SHOW_OPTION: 'SHOW_OPTION',
 });
+
+export const states = [
```
OzzieOrca, line 14: Is there a way to not hardcode this? Also aren't there other valid values like Puerto Rico and military things? 🤷‍♂️

app/views/blocks/addressQuestion.html
```diff
@@ -41,8 +41,9 @@
       ng-required="block.required"
       class="form-control"
       ng-disabled="editBlock"
+      ng-if="answer.value.country === 'US'"
     >
-      <option value="">State</option>
+      <option value="" disabled selected>State</option>
```
OzzieOrca, line 46: You should generate these from your JS array so they aren't hardcoded multiple places...

OzzieOrca (approved):
https://github.com/CruGlobal/campus-contacts-web is using a few packages to get country/region/locale data. I don't remember which package provides which but we were able to avoid hardcoding stuff:
`countrily-data`
`country-region-data`
`iso-639-1`
`m49-regions`

Also I know give-web handles international addresses by just providing 4 address lines that users can type whatever they want in. Idk if that's any better or worse than what you've got here.

app/scripts/controllers/eventDetails.js
```diff
@@ -86,6 +87,16 @@ angular
...
+    $scope.countries = allCountries;
+    $scope.conference.locationCountry = conference.locationCountry
+      ? conference.locationCountry
+      : 'US';
+    $scope.currentRegions =
+      $scope.countries[
+        $scope.countries
+          .map(c => c[1])
+          .indexOf($scope.conference.locationCountry)
+      ][2];
```
OzzieOrca, line 99: Dealing with indexes seems a little odd here. Idk if this code is perfect but seems like you could just find the country directly:
```suggestion
    $scope.currentRegions =
      $scope.countries.find(c => c[1] === $scope.conference.locationCountry)?.[2];
```

app/scripts/controllers/eventDetails.js
```diff
@@ -695,6 +706,23 @@ angular
...
+    $scope.$watch(
+      'conference.locationCountry',
+      (newVal, oldVal) => {
+        if (oldVal !== newVal) {
+          $scope.currentRegions =
+            $scope.countries[
+              $scope.countries
+                .map(c => c[1])
+                .indexOf($scope.conference.locationCountry)
+            ][2];
```
OzzieOrca, line 718: Maybe this should be a function if you're doing it multiple times.

app/scripts/directives/blockEditor.js
```diff
@@ -98,6 +100,25 @@ angular.module('confRegistrationWebApp').directive('blockEditor', function() {
...
+            zip: null,
+            country: 'US',
+          },
+        };
+      }
+
+      $scope.currentRegions = country => {
+        return $scope.countries[
+          $scope.countries.map(c => c[1]).indexOf(country)
+        ][2];
```
OzzieOrca, line 119: Another instance

app/scripts/directives/blockRegistration.js
```diff
@@ -49,6 +51,12 @@ angular
           isNew && $scope.adminEditRegistrant.answers.push($scope.answer);
         }
 
+        $scope.currentRegions = country => {
+          return $scope.countries[
+            $scope.countries.map(c => c[1]).indexOf(country)
+          ][2];
```
OzzieOrca, line 57: Another instance

app/scripts/filters/eventAddressFormat.js
```diff
@@ -0,0 +1,39 @@
...
+  .module('confRegistrationWebApp')
+  .filter('eventAddressFormat', () => (city, state, zip, country) => {
+    const formattedState =
+      state &&
+      allCountries.reduce((result, currentCountry) => {
+        if (currentCountry[1] === country) {
+          result = currentCountry[2].filter(r => r[1] === state)[0][0];
+        }
+        return result;
+      }, '');
```
OzzieOrca, line 13: Similar instance. Reduce doesn't seem appropriate here since you're searching for a single item and then operating on the single item you found.
```suggestion
      allCountries.find(currentCountry => currentCountry[1] === country)?.[2].filter(r => r[1] === state)?.[0][0];
```

app/scripts/filters/eventAddressFormat.js
```diff
@@ -0,0 +1,39 @@
...
+      : formattedState
+      ? zip
+        ? `${formattedState} ${zip}`
+        : `${formattedState}`
+      : zip
+      ? `${zip}`
+      : '';
+
+    const addressCountry = country
+      ? allCountries[allCountries.map(c => c[1]).indexOf(country)][0]
```
OzzieOrca, line 32: Another instance

app/scripts/services/validateRegistrant.js
```diff
@@ -337,10 +339,32 @@ angular
...
             if (
-              _.isEmpty(answer.address1) ||
-              _.isEmpty(answer.state) ||
-              _.isEmpty(answer.city) ||
-              _.isEmpty(answer.zip)
+              answer.country !== 'US' &&
+              (_.isEmpty(answer.address1) ||
+                _.isEmpty(answer.city) ||
+                (allCountries[
+                  allCountries.map(c => c[1]).indexOf(answer.country)
```
OzzieOrca, line 346: Another instance

app/scripts/services/validateRegistrant.js
```diff
@@ -337,10 +339,32 @@ angular
...
+                ][2].length > 1 &&
+                  !allCountries[
+                    allCountries.map(c => c[1]).indexOf(answer.country)
```
OzzieOrca, line 349: Another instance

OzzieOrca, line 349: Another instance

app/scripts/services/validateRegistrant.js
```diff
@@ -337,10 +339,32 @@ angular
...
+              invalidBlocks.push(block.id);
+              return;
+            }
+            if (
+              answer.country === 'US' &&
+              (_.isEmpty(answer.address1) ||
+                (_.isEmpty(answer.state) ||
+                  !allCountries[235][2]
+                    .map(r => r[1])
+                    .includes(answer.state)) ||
```
OzzieOrca, line 364: Another instance

app/views/paymentMethods/creditCard.html
```diff
@@ -82,7 +82,97 @@ <h4 translate>Billing Address</h4>
...
+      <div
+        class="col-sm-4 stacked-spacing-col-sm"
+        ng-show="creditCardCountry === 'US'"
+      >
+        <label translate>State</label>
+        <select
+          ng-model="currentPayment.creditCard.billingState"
+          class="form-control"
+        >
+          <option value="AL">Alabama</option>
```
OzzieOrca, line 114: Can these be pulled from the library?

OzzieOrca (changes requested):

app/scripts/controllers/eventDetails.js
TheNoodleMoose: Yeah I agree, this is a better approach.

app/views/paymentMethods/creditCard.html
TheNoodleMoose: Yeah, I forgot about this section. Updating now!

app/scripts/controllers/eventDetails.js
Benjamin-deVries: Indeed, the API's default behavior will be an empty string i.e. when the JSON is marshalled into a ConverenceEntity object on the API side, if there are fields that are omitted they will default to an empty string.

TheNoodleMoose:
@wrandall22 This is approved and on staging, just waiting to hear back from QA.

wrandall22:
@Benjamin-deVries please let me know when QA for this is done and we want to deploy.

Benjamin-deVries:
> @Benjamin-deVries please let me know when QA for this is done and we want to deploy.

@wrandall22 I believe QA is being taken care of by the products team (maybe Charles). API is working and was already deployed end of Jan. Our only remaining involvement in this was troubleshooting if UI was getting errors.

PR #780, opened 2022-01-28
EVENT-785-Hot-Fix

TheNoodleMoose:
Fix based on comments in https://jira.cru.org/browse/EVENT-785. My original logic didn't really account for group registration. Having gone through fixing this, this must have been broken since group registrations were introduced.

### DIFF
--- a/app/scripts/controllers/registration.js
+++ b/app/scripts/controllers/registration.js
@@ -37,26 +37,33 @@ angular
     $scope.savingAnswers = false;
 
     $scope.activePageId = pageId || '';
-    $scope.page = _.find(conference.registrationPages, { id: pageId });
-    $scope.activePageIndex = _.findIndex($scope.conference.registrationPages, {
-      id: pageId,
-    });
     // Filter through all pages and remove any empty pages
     // Check for registrant needed since on the welcome page
     // user could possibly not have started registration yet
     $scope.validPages =
       $scope.currentRegistration.registrants.length >= 1
         ? $scope.conference.registrationPages.filter(page => {
+            const registrantId = $scope.currentRegistrant
+              ? $scope.currentRegistrant
+              : $scope.currentRegistration.registrants[0].id;
+
             return (
-              page.blocks.filter(
-                block =>
-                  !block.registrantTypes.includes(
-                    $scope.currentRegistration.registrants[0].registrantTypeId,
+              page.blocks.filter(block =>
+                validateRegistrant.blockVisible(
+                  block,
+                  currentRegistration.registrants.find(
+                    r => r.id === registrantId,
                   ),
+                ),
               ).length > 0
             );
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/registration.js
```diff
@@ -37,26 +37,33 @@ angular
...
-    });
     // Filter through all pages and remove any empty pages
     // Check for registrant needed since on the welcome page
     // user could possibly not have started registration yet
     $scope.validPages =
       $scope.currentRegistration.registrants.length >= 1
         ? $scope.conference.registrationPages.filter(page => {
+            const registrantId = $scope.currentRegistrant
+              ? $scope.currentRegistrant
+              : $scope.currentRegistration.registrants[0].id;
```
TheNoodleMoose, line 48: When the user presses ```Go back``` on the first page, they're navigated to ```http://eventregistrationtool.com/register/eventIdHere/page/``` which means ```$scope.currentRegistrant``` is ```undefined```. So we should check if it's truthy and if not just default to the first registrant's id.

app/scripts/controllers/registration.js
```diff
@@ -37,26 +37,33 @@ angular
...
+
             return (
-              page.blocks.filter(
-                block =>
-                  !block.registrantTypes.includes(
-                    $scope.currentRegistration.registrants[0].registrantTypeId,
+              page.blocks.filter(block =>
+                validateRegistrant.blockVisible(
```
TheNoodleMoose, line 52: I should have realized there was a function that existed that was doing what I was trying to do. So I went ahead and utilized that instead.

app/scripts/controllers/reviewRegistration.js
```diff
@@ -178,19 +178,29 @@ angular
...
         .path(
           '/' +
             ($rootScope.registerMode || 'register') +
             '/' +
             conference.id +
             '/page/' +
-            conference.registrationPages[0].id,
+            conference.registrationPages.filter(page =>
+              $scope.pageIsVisible(page, registrantId),
+            )[0].id,
```
TheNoodleMoose, line 191: The same bug would occur if you clicked edit on registrant but the page was empty, so this bug is resolved now and the user is correctly navigated to the first valid page.

app/scripts/controllers/registration.js
```diff
@@ -66,6 +73,17 @@ angular
       ];
     };
 
+    $scope.getValidFirstPage = registrantId =>
```
TheNoodleMoose, line 76: We have to have a function to dynamically check what the first valid page is per registrant. We can't use ```validPages[0].id``` because the array of valid pages is generated from the current registrant's registrant type which could be different from the registrant they are trying to navigate to.

OzzieOrca (approved):

PR #781, opened 2022-01-31
EVENT-781-Improvement

TheNoodleMoose:
After talking with some of the people from the ERT team, I recommended it would be better for safety reasons to not use ```toTrustedHTML```. They do lose the ability to use CSS, but the security benefits we gain outway that utility I believe.

Security reasons behind this are the fact that ```$sce.trustAsHtml``` overrides the ```ngSanitize``` that should normally happen. This means whatever is inputted, is not checked for anything that could be 'unsafe'. The sanitize logic essentially strips out everything that is not basic HTML code, including css, scripts, etc. 

With overriding the sanitize logic, I can do the following:
Screen Shot 2022-01-31 at 10 24 37 AM
and since it's not scrubbed out, it gets executed
Screen Shot 2022-01-31 at 10 24 46 AM

### DIFF
--- a/app/views/eventOverview.html
+++ b/app/views/eventOverview.html
@@ -24,7 +24,7 @@
           <h2 class="page-title border" translate>Event Information</h2>
           <p><strong>{{conference.name}}</strong></p>
           <p
-            ng-bind-html="conference.description | toTrustedHtml"
+            ng-bind-html="conference.description"
             ng-show="conference.description"
             class="display-linebreaks"
           ></p>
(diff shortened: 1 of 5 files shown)


reldredge71 (approved):

TheNoodleMoose:
@wrandall22 Went ahead and updated the description. And after talking with the ERT team, they're not too worried about using CSS in the description. We do actually already allow for custom CSS stylesheets, so I believe they could utilize that if they ever had the desire to.

PR #784, opened 2022-02-04
NextJS

TheNoodleMoose:
Thanks to @OzzieOrca's help, we have NextJS running inside of AngularJS! 🥳

One of the main blockers was that Next.js tries to server render/pre render which doesn't work at all with angular. Scotty added the React.lazy function to wait to run any Angular code until we're on the client. That didn't work with React serverside rendering so an upgrade to the rc for v18 was needed. 

There's still some work to be done, but this is a good starting point to continue to work from and experiment with.


OzzieOrca:
@TheNoodleMoose Your changes to bootstrap in the App component weren't working for me. It seems like all AngularJS components have to be defined before bootstrapping and then all `angular2react` calls have to happen after bootstrapping.

I moved the component config into the AngularJS part of the app and changed how it bootstraps (there's more of a description in my commit message). I think this will work better as we add more pages and navigate between them. I guess each one will still need to have `import 'app/scripts/main';`. I tried moving that up to the App component with another React.lazy function but it appears to load both lazy components at the exact same which means the bootstrap hasn't happened before the `angular2react` call.

Let me know what you think. :)

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
TheNoodleMoose, line 1: @OzzieOrca I noticed that in the ```.husky``` folder there's a ```.gitignore``` that ignores all files(using ```*```). Should this have been added? I'm making a branch for the prettier changes and just want to be sure.

OzzieOrca (approved):
🎉

PR #789, opened 2022-02-11
Fix-Skip-Next-Page-Bug

TheNoodleMoose:
Interesting bug, basically a page now only renders if it has valid questions based on registrant type after my rework. But the question and answer rules also affect this. If an answer to a question causes another question to be visible on another page and the page wasn't valid before, AND that page is the next page, it basically would be skipped if you pressed 'Next Page'. But now we are continually checking valid pages as they navigate to the next page. I don't think this is needed for navigating back as rules shouldn't apply to previous pages.

### DIFF
--- a/app/scripts/controllers/registration.js
+++ b/app/scripts/controllers/registration.js
@@ -62,11 +62,33 @@ angular
             })
           : $scope.conference.registrationPages;
 
+      $scope.checkValidPages = () => {
+        $scope.validPages = $scope.conference.registrationPages.filter(
+          (page) => {
+            const registrantId = $scope.currentRegistrant
+              ? $scope.currentRegistrant
+              : $scope.currentRegistration.registrants[0].id;
+
+            return (
+              page.blocks.filter((block) =>
+                validateRegistrant.blockVisible(
+                  block,
+                  currentRegistration.registrants.find(
+                    (r) => r.id === registrantId,
+                  ),
+                ),
+              ).length > 0
+            );
+          },
+        );
+      };
+
       $scope.page = _.find($scope.validPages, { id: pageId });
       $scope.activePageIndex = _.findIndex($scope.validPages, {
         id: pageId,
       });
       $scope.nextPage = function () {
+        $scope.checkValidPages();
         var visiblePageArray = _.filter($scope.validPages, function (page) {
           return $scope.pageIsVisible(page);
         });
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/registration.js
```diff
@@ -62,11 +62,33 @@ angular
             })
           : $scope.conference.registrationPages;
 
+      $scope.checkValidPages = () => {
```
wrandall22, line 65: What calls this new function?

app/scripts/controllers/registration.js
```diff
@@ -62,11 +62,33 @@ angular
...
+          },
+        );
+      };
+
       $scope.page = _.find($scope.validPages, { id: pageId });
       $scope.activePageIndex = _.findIndex($scope.validPages, {
         id: pageId,
       });
       $scope.nextPage = function () {
+        $scope.checkValidPages();
```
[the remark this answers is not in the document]
wrandall22: Thanks! For some reason I didn't see this when I looked first haha.

wrandall22:
Seems like this new function might warrant a test.

OzzieOrca (approved):

TheNoodleMoose:
@wrandall22 This whole controller actually didn't have any tests so went ahead and created a spec for it.

test/spec/controllers/registration.spec.js
```diff
@@ -0,0 +1,110 @@
...
+  );
+
+  it('should have validPages based on current registrant', () => {
+    let validPages = scope.validPages;
+
+    expect(validPages.length).toEqual(2);
+  });
+
+  it('should getFirstValidPage', () => {
+    let current = scope.currentRegistrant;
```
wrandall22, line 46: A bit nit-picky from my, but `currentRegistrant` would be better than `current`.

test/spec/controllers/registration.spec.js
```diff
@@ -0,0 +1,110 @@
+import angular from 'angular';
```
wrandall22, line 1: Do you think it's worth putting in a test for nextPage?

test/spec/controllers/registration.spec.js
```diff
@@ -0,0 +1,122 @@
...
+    scope.checkValidPages();
+
+    expect(scope.currentRegistrant).toEqual(
+      scope.currentRegistration.registrants[1].id,
+    );
+
+    expect(scope.validPages.length).toEqual(1);
+  });
+
+  it('should run nextPage', () => {
```
wrandall22, line 111: You might confirm in here that `$scope.checkValidPages();` was called, since that is what is added in this PR.

TheNoodleMoose:
@wrandall22 thanks for the review and comments! This is on staging and working, so if everything looks good I can merge.

wrandall22 (approved):

PR #792, opened 2022-02-15
Fix-Double-Login-Modal-Bug

TheNoodleMoose:
The API recently updated so that the ```/profile``` endpoint correctly returns a ```401``` instead of a ```503```. However, this has led to a double modal being presented to the user, that persists even after they finish signing in. The ```unauthorizedInterceptor``` just checks to see if the request status was a ```401``` and if the user currently has a ```crsToken``` cookie(which most likely has expired). If so, show the modal. So when the user's token expired, they get presented a modal for the first request that was unauthorized, and then a second modal for the profile data request. I figured we don't necessarily need to have them sign in to get their profile data, as even if they were on the landing page(a page that does not require auth), the modal would pop up. Maybe this would be resolved if we removed the ```crsToken``` if we hit a ```401``` error(i.e always delete the crsToken if 401, but only show modal if the request was not for profile data)? Let me know what you think!


wrandall22:
My initial thoughts:

1. It is often hard to maintain things like `if not x endpoint` for handling logic like this (these lists tend to be always growing)
2. Is there any way for the front-end to verify if the `crsToken` is expired?
3. If we were to delete `crsToken` on `401` response, how easy is it to get a new one that is not expired?
4. I'm not quite sure I understand fully why two modals are popping up in the first place. Are we forcing a modal every time they go to get profile data regardless of the logged-in status, and before the `401` comes back?

wrandall22 (approved):

wrandall22:
I guess the part that seems to need to be smarter is this:
> So on every page load, it's requesting the data for that page, and the profile data.

The profile page should be smart enough to know that the data for that page = the profile data.

wrandall22:
Ah, I see. I guess this is where GraphQL could help. One request that gets page data + profile data. Otherwise it would be slowing down the page by requesting them sequentially. Probably lots of ways it could go to be a good solution.

PR #793, opened 2022-02-16
EVENT-791-Update-Merchant-Account-Information-Text

TheNoodleMoose:


wrandall22:
Did this wording come from somewhere or did you have to make it up yourself?

wrandall22 (approved):

TheNoodleMoose:
@wrandall22 I was provided it in https://jira.cru.org/browse/EVENT-791

wrandall22:
I am a bit curious why it is 3 translate tags instead of 1.

wrandall22:
Also, how would you go about putting the bold and italics in there? Or are they not required?

TheNoodleMoose:
@wrandall22 It's actually funny you say that. I originally pushed it with one translate tag, but then realized it had been three before and maybe it should stay that way. I went ahead and separated it and then rebased and squashed. I'm not exactly sure the context or reasoning behind it being three different translates.

wrandall22:
One possibility for italics and bold, is, if they are required, you could break up the text more into other translate tags per style. Something like:
```
  <span if-block>
    <em><translate>Italic translated text</translate></em>
    <translate>Regular text</translate>
    <strong><translate>Bold text</translate></strong>
    <translate>Regular text</translate>
```

PR #796, opened 2022-02-25
EVENT-780-Improve-Sorts-Registration-View

TheNoodleMoose:


wrandall22:
Putting our conversation onto "paper". This is basically ready from a front-end code standpoint, but waiting on the API side, right?

TheNoodleMoose:
@wrandall22 Yes, this will have to wait til https://jira.cru.org/browse/EVENT-784 is completed.

app/scripts/controllers/eventRegistrations.js
```diff
@@ -216,47 +216,25 @@ angular
...
-        } else if (orderBy === 'last_updated_timestamp') {
-          return registrant.lastUpdatedTimestamp;
-        } else if (orderBy === 'group_id') {
-          return registrant.groupId;
-        } else {
-          var answerValue = findAnswer(registrant, orderBy);
-          if (angular.isUndefined(answerValue)) {
-            return '';
-          }
+        var answerValue = findAnswer(registrant, orderBy);
```
TheNoodleMoose, line 219: I removed all of this as it was old filtering logic that the frontend was doing before API filtering was introduced. It was actually conflicting with the sorting that the API was doing.

app/views/eventRegistrations.html
```diff
@@ -463,7 +479,7 @@ <h2 class="page-title" translate>Registrations</h2>
         </thead>
         <tbody>
           <tr
-            ng-repeat-start="r in registrants | orderBy:answerSort:reversesort"
+            ng-repeat-start="r in registrants | orderBy:answerSort"
```
TheNoodleMoose, line 482: I also removed ```reverseSort``` as it was doing an additional layer of reverse sorting on the frontend, even though the API is handling that now. It was essentially undoing the reverse sorting from the API.

wrandall22 (approved):

app/scripts/controllers/eventRegistrations.js
```diff
@@ -499,36 +477,6 @@ angular
         return _.find($scope.registrations, { id: id });
       };
 
-      $scope.getRegistrationLastUpdatedTimestamp = (registrant) => {
```
TheNoodleMoose, line 502: We determined that we no longer need this function, as we are now just using the ```registrant.lastUpdatedTimestamp``` value instead.

TheNoodleMoose:
@wrandall22 If this looks good then it can be deployed whenever @Benjamin-deVries merges his changes to production.

PR #799, opened 2022-03-02
Registration-Export-Hot-Fix

TheNoodleMoose:
There have been a number of reports of issues related to registration exports recently. This is due to the rework of how the API produces the exports for registrations. For now, I'm reverting the work that will be affected by the API reverting its changes that introduced the issue/bug.


TheNoodleMoose:
@hlbraddock That would revert my conversion from angularjs -> react. When we are ready we can just revert this pr.

hlbraddock (approved):

PR #805, opened 2022-04-25
EVENT-776 - Foreign Addresses

wrandall22:
This reverts the revert of #803 and fixes the bug which caused us to revert.


wrandall22:
Basically what was happening is that old events (ones created before this code) do not necessarily have a country, so `answer.country === ''` at the point of the `allCountries.find()` call. This lead to index 2 being undefined. If we check for country being empty first, we don't run into this issue.

PR #812, opened 2022-08-01
EVENT-798 - Add chartfield to promotions

wrandall22:
[Jira](https://jira.cru.org/browse/EVENT-798) and associated https://github.com/CruGlobal/ert-api/pull/994.

This PR adds chartfield text fields to the promotions area.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist
 coverage
 package-lock.json
 .next
+.vs
--- a/app/styles/common.scss
+++ b/app/styles/common.scss
@@ -211,6 +211,14 @@ h2.page-title {
   margin-top: 10px;
 }
 
+.margin-left-0 {
+  margin-left: 0;
+}
+
+.margin-right-15 {
+  margin-right: 15px;
+}
+
 /*** Forms ***/
 
 /* Inputs */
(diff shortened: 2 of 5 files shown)


gyasistory (commented):
Is there a way to test locally this is all I get when I run it locally?

image

wrandall22:
You can test it in staging if you want. The only way to test it locally that I have found so far is to run the API locally as well, which is a bit of a hassle. This is due to the fact that the staging API Okta implementation always redirects you to the staging environment instead of back to your local.

TheNoodleMoose:
There's a way to test locally without the API running locally. It's a bit hacky though. Go through the process of signing in locally and you will then be redirected to the staging site. From here, open the dev console -> Application -> cookies -> cookies for the site. You'll want to find the `crsToken` and copy that. Now go back to localhost and navigate to route `localhost:9000/auth/pasteCrsTokenHere`. You'll now be authenticated locally! Hope this helps!
Screenshot 2022-08-01 140834

gyasistory (approved):

hlbraddock:
@wrandall22 
Note that this PR will need to be deployed in tandem with https://github.com/CruGlobal/ert-api/pull/994. 
We can manage the deploy of both, if you like.

PR #821, opened 2023-01-06
Fix inconsistencies with other radio option

canac:
Reported in [HelpScout #854613](https://secure.helpscout.net/conversation/2086060717).

**Summary of changes:**

* Getting the logic right for the answer value with two-way data binding based on two sources of data is extremely tricky, if not impossible, so I refactored it to use two separate variables. The first tracks the selected answer and the second tracks the text of the other input. Those two variables are then combined to produce the answer value.
* Validation in the edit registration modal is now performed in the same way as in the add registration modal.

### DIFF
--- a/app/scripts/directives/blocks.js
+++ b/app/scripts/directives/blocks.js
@@ -89,26 +89,40 @@ angular
       templateUrl: radioQuestionTemplate,
       restrict: 'E',
       controller: function ($scope) {
-        $scope.$watch('answer.value', function () {
-          if (angular.isDefined($scope.answer)) {
-            //check if answer is not in current choices
-            if (
-              !_.includes(
-                _.map($scope.block.content.choices, 'value'),
-                $scope.answer.value,
-              )
-            ) {
-              $scope.otherAnswer = $scope.answer.value;
-            }
+        $scope.otherSentinel = '__other__';
+
+        $scope.selectOtherAnswer = () => {
+          if ($scope.otherAnswer) {
+            $scope.selectedAnswer = $scope.otherSentinel;
+          }
+        };
+
+        $scope.clearAnswer = () => {
+          $scope.selectedAnswer = '';
+        };
+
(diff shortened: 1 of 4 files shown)


app/views/blocks/radioQuestion.html
```diff
@@ -15,7 +15,7 @@
       show-errors
       name="{{ block.id }}"
       ng-required="block.required"
-      ng-model="answer.value"
+      ng-model="$parent.selectedAnswer"
```
wrandall22, line 18: Do you actually need to specify `$parent`?

canac: Yes, otherwise it tries to read/write the `selectedAnswer` property on the new scope created by `ng-repeat`.

wrandall22:
Feel free to move it to stage to test it there.

canac:
Just tested in staging, and it seems to work.

app/scripts/directives/blocks.js
```diff
@@ -89,26 +89,40 @@ angular
...
-            return;
+        $scope.selectOtherAnswer = () => {
+          if ($scope.otherAnswer) {
+            $scope.selectedAnswer = $scope.otherSentinel;
           }
-          $scope.answer.value = $scope.otherAnswer;
         };
+
+        $scope.clearAnswer = () => {
+          $scope.selectedAnswer = '';
```
wrandall22, line 101: I noticed this line isn't covered by tests. Did you get to test the case where you type in an "Other" answer, submit the registration, edit the registration, clear the Other answer, and submit again? I think that was the edge case I didn't get working before, basically I could submit an empty other answer.

canac: I can add tests to cover the missing lines, but the Other input is required when the other option is selected, so it doesn't let you clear the other answer and submit.

wrandall22 (approved):

PR #823, opened 2023-01-09
Debounce option changes

canac:
For some reason, `ng-model-options="{ debounce: 1000 }"` doesn't play nicely with radio buttons. If you click one option then a different one within one second, the selection is cleared when the debounce finally updates the value, even though the modal value is still set. Someone had a similar issue [here](https://stackoverflow.com/questions/31254291/angularjs-debounce-is-clearing-my-radio-input) with no resolution besides using a different method for debouncing.

Work around this issue by manually performing the debouncing within the controller.

### DIFF
--- a/app/scripts/directives/blocks.js
+++ b/app/scripts/directives/blocks.js
@@ -88,7 +88,7 @@ angular
     return {
       templateUrl: radioQuestionTemplate,
       restrict: 'E',
-      controller: function ($scope) {
+      controller: function ($scope, $timeout) {
         $scope.otherSentinel = '__other__';
 
         $scope.selectOtherAnswer = () => {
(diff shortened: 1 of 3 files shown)


wrandall22:
I wonder if the debounce was only there for the Other option. Is it still worth having with the refactoring that was done?

canac:
@wrandall22 All other input types have `ng-model-options="{ debounce: 1000 }"` on the inputs. I'm not sure what the reasoning for the debouncing is, but keeping the debouncing for the radio buttons keeps it consistent with the other input types.

app/scripts/directives/blocks.js
```diff
@@ -116,14 +116,29 @@ angular
           $scope.otherAnswer = '';
         }
 
-        $scope.$watchGroup(['selectedAnswer', 'otherAnswer'], () => {
-          if ($scope.answer) {
-            $scope.answer.value =
-              $scope.selectedAnswer === $scope.otherSentinel
-                ? $scope.otherAnswer
-                : $scope.selectedAnswer;
+        // Manually implement debounce so that $timeout can be mocked in tests
```
wrandall22, line 119: A bit nitpicky, but:
```suggestion
        // Manually implement debounce in a way that $timeout can be mocked in tests
```
or

```suggestion
        // Manually implement debounce because debounce does not play well with radio buttons
```

We aren't manually implementing because we want `$timeout` to be mocked, but because debounce is not working well for radio buttons.

canac: Good point. I initially meant that I was writing a custom debouncer instead of using `_.debounce` because I needed it to use `$timeout`, but that's definitely not clear.

wrandall22 (approved):

PR #824, opened 2023-01-10
Fix file is undefined exception

canac:
Fixes this Rollbar exception caused by `files` being an empty array: https://rollbar.com/Cru/ert-web/items/601/occurrences/307996597668/


wrandall22 (approved):

canac:
@wrandall22 Removed that blank line. Is it good now? The only other change since your last review is removing the `$timeout` wrapper. It is unnecessary because the code is already in a digest cycle because it is the resolve handler of `$q` promise.

PR #825, opened 2023-01-13
[EVENT-800] Support promo report upload

canac:
https://jira.cru.org/browse/EVENT-800

Notes:
* ~~I'd like to add unit tests, at least for the logic in the new hooks, but that will require setting up a React testing library, and I wanted to go ahead and start getting feedback on the actual code~~
**Edit:** I added tests for the most complex and bug-prone code paths.
* I rewrote the journal upload page in React in the process of implementing this feature, but I'm going to clean it up and commit in a separate PR to minimize the risk of breaking something


app/scripts/components/RegistrationsPaidPopover/RegistrationsPaidPopover.tsx
```diff
@@ -0,0 +1,9 @@
+export const RegistrationsPaidPopover = (): JSX.Element => {
+  return (
+    <div>
+      <button className="btn btn-default btn-xs">Balance Due</button>{' '}
```
wrandall22, line 4: I wonder if `&nbsp;` is better than a space like this? Or perhaps better still is to use CSS to separate these instead.

canac: This is mostly a prettier artifact. prettier wants to break up the line to keep it from being too long, but since whitespace is trimmed at the end of a line, it has to use `{' '}` to force a space. But I can switch to using CSS to separate the buttons.

types/promotion.ts
```diff
@@ -0,0 +1,18 @@
+export interface Promotion {
+  id: string;
+  conferenceId: string;
+  registrantTypeIds: string[];
```
dr-bizz, line 4: I think the preferred way to write `string[]` is like the below? Although, I so like the ease of `string[]`
```suggestion
registrantTypeIds: Array<string>;```

File: app/scripts/hooks/usePromoReport.ts [hunk 3]
```diff
@@ -0,0 +1,101 @@
...
+    } else {
+      // To avoid getting confused if there are multiple requests in progress, track the most recent
+      // request and ignore the response if a request comes back that isn't the most recent request.
+      const currentRequest = promoReportService.loadPromoReport(
+        conference.id,
+        reportId,
+      );
+      activeRequest = currentRequest;
+      currentRequest.then((report) => {
+        if (activeRequest === currentRequest) {
```
dr-bizz, line 51: Won't `activeRequest` always equal `currentRequest` as the variable `activeRequest` is set to `currentRequest` 2 lines up?

canac: No, because it's defined outside of the hook. The idea is that `activeRequest` will keep being overwritten by each request and will hold the most recent request, while `currentRequest` will remain constant. After a request asynchronously finishes, it can check whether a different request started after it, in which case its response will be discarded in favor of the more recent request. It's a complex flow, so sorry if my explanation doesn't make sense. I'm also in the process of switching from a global `activeRequest` variable to a React ref so that multiple active `usePromoReport` hooks won't overwrite each others' `activeRequest`.

app/scripts/components/FormStatusModal/FormStatusModal.tsx
```diff
@@ -71,50 +71,42 @@ const FormStatusModal = ({
         >
           &times;
         </button>
-        <h4 translate="yes">Form Status</h4>
+        <h4>Form Status</h4>
```
dr-bizz, line 74: I assume you wanted to take the `translate` attribute off these elements, but I just want to double-check.

app/scripts/components/RegistrationFilters/RegistrationFilters.tsx
```diff
@@ -0,0 +1,283 @@
...
+  showPagination: boolean;
+  pageCount: number;
+  children: React.ReactFragment;
+  hiddenFilters?: Array<keyof RegistrationQueryParams>;
+}
+
+export const RegistrationFilters = (
+  props: RegistrationFiltersProps,
+): JSX.Element => {
+  const [queryParameters, setQueryParameters] = useState(
```
dr-bizz, line 21: Might be worth to define the type here

app/scripts/components/RegistrationFilters/RegistrationFilters.tsx
```diff
@@ -0,0 +1,283 @@
...
+            </div>
+          )}
+          <div className="more-filters col-md-2 col-sm-6 text-right stacked-spacing-col-md">
+            <br />
+            <Button
+              className="btn btn-default"
+              active={showMoreFilters}
+              onClick={() => setShowMoreFilters(!showMoreFilters)}
+            >
+              Filters{' '}
```
dr-bizz, line 170: You could add styles to the `<i>` is you need space between the text and the icon.

app/scripts/components/FormStatusModal/FormStatusModal.tsx
canac: Correct. It doesn't do anything, we aren't doing label translation in this project, and it was causing a distracting JSX error for each occurrence.

canac:
@dr-bizz I have some TypeScript config to fix, but I added the tests (not full test-coverage but at least the most complex and bug-prone parts) and addressed your concerns if you want to look this over again today or Monday.

app/scripts/components/PromoTransactionsTable/PromoTransactionsTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+                promoRegistration.registration.primaryRegistrantId,
+          )
+          .map((registrant) => ({
+            id: `${registrant.id}|${promoRegistration.promotion.id}`,
+            promoRegistration,
+            registrant,
+            ...promoRegistration,
+          })),
+      ),
+    [promoRegistrations],
```
dr-bizz, line 83: This shouldn't stop this from going live, but something to fix at a later date.

On re-renders, React is caching a new version of the `promoRegistrations` array in PromoUploadPage.tsx, which will trigger this `useMemo` to run the calulation again with the new `promoRegistrations` as it's a dependency.

The way to fix it is to add an `useMemo` to the creation of `promoRegistrations`, so it only runs the function again if data in it's dependency changes.

canac: I just fixed what I believe you're referring to.

dr-bizz (commented):
Looks good

dr-bizz (approved):

PR #826, opened 2023-01-18
Allow editing incomplete registrations

canac:
Follow-up to #821. Now, incomplete registrations can be edited again, even when required are left blank. The added `novalidate` attribute keeps the browser from showing native UI for blank required fields.

### DIFF
--- a/app/views/modals/editRegistration.html
+++ b/app/views/modals/editRegistration.html
@@ -1,4 +1,4 @@
-<form class="crs-form" name="editRegistrationForm">
+<form class="crs-form" name="editRegistrationForm" novalidate>
   <div class="modal-header">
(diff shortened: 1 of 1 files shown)


wrandall22:
So, this allows the `Submit` button, but not the `Save & Mark Complete` button to be pressed when required fields are not set?

canac:
Correct. There was a HelpScout issue from someone who wanted to edit fields on an incomplete registration like they used to be able to. It was in a HelpScout inbox that I don't have access to, so I don't have the link to it. I have confirmed with stakeholders that validation should run for "Save & Mark Complete", but not for "Save".

wrandall22 (approved):

PR #830, opened 2023-04-03
Add client-side email address validation

canac:
Implement client-side email address validation and show server-side validation errors.

### DIFF
--- a/app/views/eventDetails.html
+++ b/app/views/eventDetails.html
@@ -25,7 +25,7 @@
           type="submit"
           class="btn btn-success btn-lg"
           ng-click="saveEvent()"
-          ng-disabled="eventDetails.$invalid"
+          ng-disabled="!eventDetails.$valid"
           translate
         >
           Save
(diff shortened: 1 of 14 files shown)


app/views/eventDetails.html
```diff
@@ -25,6 +25,7 @@
           type="submit"
           class="btn btn-success btn-lg"
           ng-click="saveEvent()"
+          ng-disabled="eventDetails.$invalid"
```
dr-bizz, line 28: You should be fine with using `eventDetails.$invalid` as there isn't any async functions for saving the data, but it might be best to use `!eventDetails.$valid` to keep it simular with the current code.

In case you didn;t know, here's an article which explains the difference between invalid and valid - https://itnext.io/valid-and-invalid-in-angular-forms-61cfa3f2a0cd

app/views/eventPermissions.html
```diff
@@ -137,12 +147,13 @@ <h3 translate>Add User</h3>
           type="button"
           class="btn btn-success btn-block"
           ng-click="addPermission(addPermissionsEmail, addPermissionsLevel)"
+          ng-disabled="permissions.$invalid"
```
dr-bizz, line 150: Might be best to use `permissions.$valid` as when the form is in pending state, `permissions.$invalid` is set to FALSE.

dr-bizz (approved):

PR #831, opened 2023-05-08
[no-Jira] Remove 100 item page option

canac:
We are occasionally running into timeouts when uploading 100 transactions. Temporarily prevent these errors by making 50 the most transactions that can be viewed at one time. This will be reverted in the future when we have time for a better solution, like implementing batching on the server.

https://secure.helpscout.net/conversation/2211009504/924915/

Tested in staging ✅


dr-bizz (approved):
Can you leave a note in the code for future developers?

PR #832, opened 2023-05-17
[HS-946275] Fix multiple Angular directives error

canac:
Fix `Multiple directives [pickADate, showErrors (module: confRegistrationWebApp)] asking for new/isolated scope on: <pick-a-date>` error introduced in 192aaa30cd1c95ca990c815cccb36c19f55c0ab7.

Solution: revert the `show-errors` directive back to no longer creating a new isolated scope and instead using `$scope.$eval` to determine the value of the `show-errors-instant` attribute.

https://secure.helpscout.net/conversation/2245104032/946275?folderId=7599852

### DIFF
--- a/app/scripts/directives/showErrors.js
+++ b/app/scripts/directives/showErrors.js
@@ -2,9 +2,6 @@ angular.module('confRegistrationWebApp').directive('showErrors', function () {
   return {
     restrict: 'A',
     require: 'ngModel',
-    scope: {
-      showErrorsInstant: '<',
-    },
     link: function (scope, element, attrs, ngModelCtrl) {
       //Logic to handle groups of inputs in one block
       if (attrs.showErrors === 'group') {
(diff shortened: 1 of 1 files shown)


dr-bizz (approved):

app/scripts/directives/showErrors.js
```diff
@@ -20,7 +17,7 @@ angular.module('confRegistrationWebApp').directive('showErrors', function () {
             ngModelCtrl.$invalid &&
             (scope.currentPageVisited ||
               ngModelCtrl.$touched ||
-              !!scope.showErrorsInstant)
+              !!scope.$eval(attrs.showErrorsInstant))
```
wrandall22, line 20: Before, it used `!!attrs.showErrorsInstant`. I'm curious why it needed to change to `!!scope.$eval(attrs.showErrorsInstant)`. Is something processing differently on edit?

canac: Before, any non-empty value was interpreted as `true`. However, I needed `show-errors-instant` to be able to be an expression for this line https://github.com/CruGlobal/conf-registration-web/blob/d22b81295a235aca70d8989780864113dc054b53/app/views/blocks/emailQuestion.html#L3 so that previously saved invalid emails would be immediately marked as errors, as requested by Ben.

PR #835, opened 2023-07-26
EVENT-821 Add Event Type dropdown menu

caleballdrin:
## Description

https://jira.cru.org/browse/EVENT-821

On the Details tab of an event ERT we need to be able to collect the event type in order to capture the participant journey and know the event journey the Cru participant has followed during their time with Cru.  While the list is Campus specific the same process could be used for other ministries based on the participant event journey they with to follow.

 

When an admin answers "Yes" this is a Cru event we can use the "Ministry Hosting Event" dropdown to determine what event types would be needed. 

In this case if the Ministry Hosting the event as “Campus - Field Team/Strategy” or “Campus - National Team/Strategy” and I would also suggest that when the Ministry Purpose is selected as “Ministry Conference/Event” or "Ministry Mission Trip".  The following list should also be added to the right of the Ministry Purpose box with a tile "Event Type". 

Event Type
Fall Retreat/Getaway
Mobilization/Recruiting
Spring Break
Winter Conference
N/A
 

Be mindful to not break existing links to the Liability form process that currently uses the "Ministry Hosting Event" and "Ministry Purpose" dropdowns. 

## Changes I made

Companion PR https://github.com/CruGlobal/ert-api/pull/1029

### DIFF
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -485,6 +485,18 @@ angular
           validationErrorsHint = eventInformationPageHint;
         }
 
+        if (
+          $scope.conference.cruEvent &&
+          $scope.conference.ministry &&
+          $scope.getEventTypes().length !== 0 &&
+          !$scope.conference.eventType
+        ) {
+          validationErrors.push(
+            'Please enter which Event Type is applicable for this event.*',
+          );
+          validationErrorsHint = eventInformationPageHint;
+        }
+
         if ($scope.conference.cruEvent && !$scope.conference.type) {
           validationErrors.push('Please enter Ministry Purpose.*');
           validationErrorsHint = eventInformationPageHint;
@@ -655,11 +667,27 @@ angular
         return currentMinistry ? currentMinistry.activities : [];
       };
 
+      $scope.getEventTypes = () => {
+        const currentMinistry =
+          $scope.ministries &&
+          $scope.ministries.find((m) => m.id === $scope.conference.ministry);
+        const currentPurpose =
+          $scope.ministryPurposes &&
+          $scope.ministryPurposes.find((p) => p.id === $scope.conference.type);
+        return $scope.eventTypes &&
+          (currentMinistry?.name?.includes('Campus') ||
+            currentPurpose?.name?.includes('Mission') ||
+            currentPurpose?.name?.includes('Conference'))
+          ? $scope.eventTypes
+          : [];
+      };
+
       $scope.$watch(
         'conference.ministry',
         function (newVal, oldVal) {
           if (newVal !== oldVal) {
             $scope.conference.strategy = null;
+            $scope.conference.eventType = null;
             $scope.conference.ministryActivity = null;
           }
         },
@@ -673,6 +701,7 @@ angular
             $scope.conference.ministry = null;
             $scope.conference.strategy = null;
             $scope.conference.type = null;
+            $scope.conference.eventType = null;
             $scope.conference.ministryActivity = null;
             $scope.conference.workProject = false;
           }
@@ -735,6 +764,13 @@ angular
       $http({
         method: 'GET',
         url: 'types',
+      }).then(function (response) {
+        $scope.ministryPurposes = response.data;
+      });
+
+      $http({
+        method: 'GET',
+        url: 'event/types',
       }).then(function (response) {
         $scope.eventTypes = response.data;
       });
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -108,14 +108,12 @@ angular
               });
               return {
                 id: existingChild ? existingChild.id : uuid(),
-                name: t.name,
+                //name: t.name,
                 childRegistrantTypeId: t.id,
                 numberOfChildRegistrants: existingChild
                   ? existingChild.numberOfChildRegistrants
                   : 0,
-                selected:
-                  existingChild !== undefined &&
-                  existingChild.selected !== false,
+                //selected: existingChild !== undefined && existingChild.selected !== false,
               };
             },
           );
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -675,9 +675,13 @@ angular
           $scope.ministryPurposes &&
           $scope.ministryPurposes.find((p) => p.id === $scope.conference.type);
         return $scope.eventTypes &&
-          (currentMinistry?.name?.includes('Campus') ||
-            currentPurpose?.name?.includes('Mission') ||
-            currentPurpose?.name?.includes('Conference'))
+          ((currentMinistry &&
+            currentMinistry.name &&
+            currentMinistry.name.includes('Campus')) ||
+            (currentPurpose &&
+              ((currentPurpose.name &&
+                currentPurpose.name.includes('Mission')) ||
+                currentPurpose.name.includes('Conference'))))
           ? $scope.eventTypes
           : [];
       };
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -480,7 +480,7 @@ angular
           !$scope.conference.ministryActivity
         ) {
           validationErrors.push(
-            'Please enter which Ministry Activitiy is applicable for this event.*',
+            'Please enter which Ministry Activity is applicable for this event.*',
           );
           validationErrorsHint = eventInformationPageHint;
         }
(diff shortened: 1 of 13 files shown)


app/scripts/controllers/eventDetails.js
```diff
@@ -655,11 +667,27 @@ angular
...
+          $scope.ministryPurposes &&
+          $scope.ministryPurposes.find((p) => p.id === $scope.conference.type);
+        return $scope.eventTypes &&
+          (currentMinistry?.name?.includes('Campus') ||
+            currentPurpose?.name?.includes('Mission') ||
+            currentPurpose?.name?.includes('Conference'))
+          ? $scope.eventTypes
+          : [];
+      };
+
```
dr-bizz, line 684: This is causing the tests to fail

app/scripts/controllers/eventDetails.js
```diff
@@ -485,6 +485,18 @@ angular
...
         }
 
+        if (
+          $scope.conference.cruEvent &&
+          $scope.conference.ministry &&
+          $scope.getEventTypes().length !== 0 &&
+          !$scope.conference.eventType
+        ) {
+          validationErrors.push(
+            'Please enter which Event Type is applicable for this event.*',
```
wrandall22, line 495: Is this `*` at the end necessary?

app/views/eventDetails/eventInformation.html
```diff
@@ -288,7 +288,20 @@ <h2 class="page-title border" translate>Event Information</h2>
     <select
       class="form-control"
       ng-model="conference.type"
-      ng-options="type.id as type.name for type in eventTypes | orderBy:'name':false:sortNamesWithNA"
+      ng-options="type.id as type.name for type in ministryPurposes | orderBy:'name':false:sortNamesWithNA"
```
wrandall22, line 291: If changing the name of `eventTypes` here we should probably also change `type` for `purpose`

app/scripts/controllers/eventDetails.js
caleballdrin: Screenshot 2023-07-28 at 11 36 23 AM
The other error messages have them too.

app/views/eventDetails/eventInformation.html
caleballdrin: I think that 'type' is used by the API for ministry purpose. So I didn't want to break anything... We are adding eventType as a new field.

caleballdrin: @wrandall22 @dr-bizz Are you able to see any reason why my tests are failing?

dr-bizz: Looking into it

app/scripts/controllers/eventDetails.js
```diff
@@ -108,14 +108,12 @@ angular
               });
               return {
                 id: existingChild ? existingChild.id : uuid(),
-                name: t.name,
+                //name: t.name,
```
dr-bizz, line 111: Why is this commented out. Are we no longer needing `name` or `selected`?

app/scripts/controllers/eventDetails.js
```diff
@@ -655,11 +665,27 @@ angular
...
+          $scope.ministries.find((m) => m.id === $scope.conference.ministry);
+        const currentPurpose =
+          $scope.ministryPurposes &&
+          $scope.ministryPurposes.find((m) => m.id === $scope.conference.type);
+        return currentMinistry.eventTypes &&
+          currentPurpose &&
+          ((currentPurpose.name && currentPurpose.name.includes('Mission')) ||
+            currentPurpose.name.includes('Conference'))
+          ? currentMinistry.eventTypes
+          : [];
```
dr-bizz, line 680: I'm unsure if this is correct. Currently the event type field is only showing when a `Ministry Hosting Event` selected which has `eventTypes` and a `Ministry Purpose` was selected

app/scripts/controllers/eventDetails.js
```diff
@@ -655,11 +665,27 @@ angular
...
+      };
+
       $scope.$watch(
         'conference.ministry',
         function (newVal, oldVal) {
           if (newVal !== oldVal) {
             $scope.conference.strategy = null;
+            $scope.conference.eventType = null;
```
dr-bizz, line 688: You probably want to add a `$watch` to the ministry Purpose field, in case the admin changes the value, we would want the event type field to disappear.

test/spec/controllers/eventDetails.spec.js
```diff
@@ -27,212 +27,252 @@ describe('Controller: paymentModal', function () {
...
-    expect(scope.conference.registrationPages[0].blocks[2].tag).toEqual(null);
+    it('saveEvent() should validate the Ministry Purpose', () => {
+      scope.saveEvent();
 
-    expect(scope.conference.registrationPages[0].blocks[2].tag).toEqual(null);
+      expect(scope.notify.message.toString()).toContain(
+        'Please enter Ministry Purpose.',
+      );
+    });
   });
```
dr-bizz, line 277: Can you do a test to show what if a Ministry purpose wasn't selected but a hosting event was?

test/spec/controllers/eventDetails.spec.js
```diff
@@ -27,212 +27,252 @@ describe('Controller: paymentModal', function () {
...
 
-    expect(scope.conference.registrationPages[0].blocks[2].tag).toEqual(null);
+    it('saveEvent() should validate the Ministry Purpose', () => {
+      scope.saveEvent();
 
-    expect(scope.conference.registrationPages[0].blocks[2].tag).toEqual(null);
+      expect(scope.notify.message.toString()).toContain(
+        'Please enter Ministry Purpose.',
+      );
+    });
```
dr-bizz, line 276: Also another test when it's not a Cru event, that it shouldn't show any errors about event type?

test/spec/testData.spec.js
```diff
@@ -507,6 +508,10 @@ angular.module('confRegistrationWebApp').service('testData', function () {
     accountNumber: null,
     glAccount: null,
     cruEvent: true,
+    ministry: 'f6d31fe3-7078-4fac-a37b-9596d57558e9',
+    eventType: null,
+    type: 'ef4ffa14-0b02-4d7e-915e-77ccf958f5b9',
+    //eventType: 'cfc2b308-566b-432b-bee4-4ed60fec5608',
```
dr-bizz, line 514: Remove line

test/spec/testData.spec.js
```diff
@@ -1495,4 +1500,66 @@ angular.module('confRegistrationWebApp').service('testData', function () {
...
+  //     id: 'cfc2b308-566b-432b-bee4-4ed60fec5608',
+  //     name: 'Fall Retreat/Getaway',
+  //     $hashKey: 'object:694',
+  //   },
+  //   {
+  //     id: 'b281a0f3-b0bc-4f8e-9f49-0c3adfe25584',
+  //     name: 'Mobilization/Recruiting',
+  //     $hashKey: 'object:695',
+  //   },
+  // ];
```
dr-bizz, line 1564: remove lines

dr-bizz (changes requested):
There needs to be some changes but overall it looks good.

app/scripts/controllers/eventDetails.js
```diff
@@ -480,7 +478,19 @@ angular
...
-            'Please enter which Ministry Activitiy is applicable for this event.*',
+            'Please enter which Ministry Activity is applicable for this event.*',
+          );
+          validationErrorsHint = eventInformationPageHint;
+        }
+
+        if (
+          $scope.conference.cruEvent &&
+          $scope.conference.ministry &&
+          $scope.getEventTypes().length !== 0 &&
```
dr-bizz, line 489:
```suggestion
          $scope.getEventTypes().length &&
```

app/views/eventDetails/eventInformation.html
```diff
@@ -288,7 +288,20 @@ <h2 class="page-title border" translate>Event Information</h2>
...
+    >
+      <option></option>
+    </select>
+  </div>
+  <div
+    class="col-sm-5"
+    ng-if="conference.cruEvent && getEventTypes().length != 0"
```
dr-bizz, line 298:
```suggestion
    ng-if="conference.cruEvent && getEventTypes().length"
```

dr-bizz, line 298:
```suggestion
    ng-if="conference.cruEvent && getEventTypes().length"
```

dr-bizz (changes requested):

app/scripts/controllers/eventDetails.js
caleballdrin: We had a convo about this in the comments of the Jira ticket so now it only shows when certain Ministry Purposes are also selected.

app/scripts/controllers/eventDetails.js
caleballdrin: Ah good catch. I'll add that.

test/spec/controllers/eventDetails.spec.js
caleballdrin: got it

app/scripts/controllers/eventDetails.js
caleballdrin: I was planning to leave it commented out instead of removing it just in case we need to look back at it.

wrandall22: That's what GitHub history is for :)

dr-bizz: Let's remove the line instead of commenting it out.

app/scripts/controllers/eventDetails.js
```diff
@@ -672,14 +672,25 @@ angular
         const currentPurpose =
           $scope.ministryPurposes &&
           $scope.ministryPurposes.find((m) => m.id === $scope.conference.type);
-        return currentMinistry.eventTypes &&
+        return currentMinistry &&
```
dr-bizz, line 673: You can use optional chaining for `currentMinistry?.eventTypes` since you're not using `.includes()` on it.

```suggestion
        return currentMinistry?.eventTypes &&
```

dr-bizz (changes requested):
Just one slight change

dr-bizz (approved):
Thanks for getting this done

PR #836, opened 2023-07-26
EVENT-822 Change Gender to Sex

caleballdrin:
## Description
https://jira.cru.org/browse/EVENT-822

With the Salesforce integration we need to encourage the use of the Cru Profile questions to ensure better data and ensure our constituents are known by Cru.  With the release of the Cru Compassionate and Faithful material we have discussed with the cru theology team the need for an update to the current "Gender" Cru profile question so that we can continue to capture a binary "Male" "Female" answer.  

The current Cru Profile and default question needs to be updated from "Gender" to "Sex"

##Changes made
- Change UI strings
- Add a conditional to replace GENDER with Sex in the error message when multiple blocks are selected as the 'Cru Profile'


hlbraddock:
I'll give informal approval, but will defer official PR approval to @dr-bizz

app/scripts/directives/blockEditor.js
```diff
@@ -325,11 +325,19 @@ angular.module('confRegistrationWebApp').directive('blockEditor', function () {
             });
           });
           if (profileCount > 1) {
-            modalMessage.error(
-              'Only one ' +
+            if ($scope.block.profileType == 'GENDER') {
+              var pType = 'Sex';
+            } else {
+              pType =
```
dr-bizz, line 331: `pType` isn't defined before you use it. it's defined in the above If statement, which wouldn't work on the modalMessage. It needed to be defined outside the if statement, so it can be used in the If statement and on modalMessage.

dr-bizz: After testing, I found that what you have works. But in Typescript (TS), this would throw an error. Typescript is a stricter version of JS, so when possible you want to write like TS when writing JS.

dr-bizz (changes requested):

app/scripts/directives/blockEditor.js
caleballdrin: Ok I changed it to const like you suggested.

app/scripts/directives/blockEditor.js
```diff
@@ -324,12 +324,20 @@ angular.module('confRegistrationWebApp').directive('blockEditor', function () {
               }
             });
           });
+
```
dr-bizz, line 327: Was this new line added on purpose?

app/scripts/directives/blockEditor.js
```diff
@@ -324,12 +324,20 @@ angular.module('confRegistrationWebApp').directive('blockEditor', function () {
...
+            const pType =
+              $scope.block.profileType == 'GENDER'
+                ? 'Sex'
+                : $scope.block.profileType.charAt(0).toUpperCase() +
+                  $scope.block.profileType
+                    .split('_')
+                    .join(' ')
+                    .slice(1)
+                    .toLowerCase();
+
```
dr-bizz, line 338: Wa this new line added on purpose too?

caleballdrin: Ah... no it was not

PR #837, opened 2023-08-02
EVENT-825 Campus Question Force Selection Option

caleballdrin:
## Description
https://jira.cru.org/browse/EVENT-825

With the Salesforce integration we need to encourage the use of the Cru Profile questions to ensure better data and ensure our constituents are known by Cru.  The Current "Campus" default question pulls from a database of Campuses but does require the registrant to select anything on the drop down list.  This leads to inconsistent data as the same school could be represented in many ways with unlimited free form response.

The proposed improvement is as follows:

Update the default name when the question is used: "Campus- Type the full name of your campus and choose from the list that appears”

## Changes I made

- Stopped accepting user input that does not match the campus database with _typeahead-editable="false"_
- Changed debounce and wait time to 100ms so the list of options pops up sooner.
- Added error message when a users input is not found in the campus list.
- Add a default ExportFieldTitle and set it to "Campus".
- Change the default title of the Campus Question
- Add an admin checkbox to allow the admin to show International Campuses in the autocomplete.
- Set a limit on the number of campuses that show up in the autocomplete. Set a maxheight.

Companion PR https://github.com/CruGlobal/ert-api/pull/1030

### DIFF
--- a/app/scripts/directives/questionToolbar.js
+++ b/app/scripts/directives/questionToolbar.js
@@ -175,6 +175,7 @@ angular
             defaultProfile: 'CAMPUS',
             iconClass: 'fa-graduation-cap',
             name: 'Campus',
+            defaultExportFieldTitle: 'Campus',
           },
           {
             id: 'dormitoryQuestion',
--- a/app/scripts/directives/questionToolbar.js
+++ b/app/scripts/directives/questionToolbar.js
@@ -112,7 +112,7 @@ angular
           {
             id: 'numberQuestion',
             defaultTitle: 'Number',
-            iconClass: 'fa-superscript',
+            iconClass: 'fa-calculator',
             name: 'Number',
           },
           {
@@ -173,7 +173,7 @@ angular
             defaultTitle:
               'Campus- Type the full name of your campus and choose from the list that appears',
             defaultProfile: 'CAMPUS',
-            iconClass: 'fa-graduation-cap',
+            iconClass: 'fa-university',
             name: 'Campus',
             defaultExportFieldTitle: 'Campus',
           },
@@ -181,7 +181,7 @@ angular
             id: 'dormitoryQuestion',
             defaultTitle: 'Dormitory',
             defaultProfile: 'DORMITORY',
-            iconClass: 'fa-graduation-cap',
+            iconClass: 'fa-building',
             name: 'Dormitory',
           },
         ];
--- a/app/scripts/directives/questionToolbar.js
+++ b/app/scripts/directives/questionToolbar.js
@@ -149,10 +149,10 @@ angular
           },
           {
             id: 'genderQuestion',
-            defaultTitle: 'Gender',
+            defaultTitle: 'Sex',
             defaultProfile: 'GENDER',
             iconClass: 'fa-male',
-            name: 'Gender',
+            name: 'Sex',
           },
           {
             id: 'yearInSchoolQuestion',
--- a/app/scripts/directives/questionToolbar.js
+++ b/app/scripts/directives/questionToolbar.js
@@ -171,7 +171,7 @@ angular
           {
             id: 'campusQuestion',
             defaultTitle:
-              'Campus- Type the full name of your campus and choose from the list that appears',
+              'Campus - Type the full name of your campus and choose from the list that appears',
             defaultProfile: 'CAMPUS',
             iconClass: 'fa-university',
             name: 'Campus',
--- a/app/scripts/directives/questionToolbar.js
+++ b/app/scripts/directives/questionToolbar.js
@@ -170,7 +170,8 @@ angular
           },
           {
             id: 'campusQuestion',
-            defaultTitle: 'Campus',
+            defaultTitle:
+              'Campus- Type the full name of your campus and choose from the list that appears',
             defaultProfile: 'CAMPUS',
             iconClass: 'fa-graduation-cap',
             name: 'Campus',
--- a/app/views/blocks/campusQuestion.html
+++ b/app/views/blocks/campusQuestion.html
@@ -2,14 +2,18 @@
   <input
     type="text"
     ng-model="answer.value"
-    ng-model-options="{ debounce: 1000 }"
+    ng-model-options="{ debounce: 100 }"
     placeholder=""
     class="form-control"
+    show-errors
     ng-required="block.required"
     uib-typeahead="campus for campus in searchCampuses($viewValue)"
     typeahead-loading="loadingCampuses"
-    typeahead-min-length="4"
-    typeahead-wait-ms="300"
+    typeahead-min-length="2"
+    typeahead-wait-ms="100"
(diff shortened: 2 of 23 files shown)


dr-bizz:
In the description, it mentions adding an additional checkbox Option on the admin panel that would “Force Selection from Cru Database”. I can't see that in this PR.

dr-bizz (changes requested):
The code looks good. There seems to be a few things missing on this PR.

The text added to the field when used. (I've mentioned this in another comment).

There isn't a checkbox on the admin panel to allow the admin to Force Selection from Cru Database

dr-bizz:
It's worth making the campus dropdown have a `max-height` so prevent the from having to scroll down the page to see all the options when it returns lots.

caleballdrin:
> 

I was looking into setting max-height and came across this issue. Maybe we should just limit the results to 10 from the API.
https://github.com/ng-bootstrap/ng-bootstrap/issues/1119

dr-bizz:
Seems like there is a way but you have to make it work - https://github.com/ng-bootstrap/ng-bootstrap/issues/1119#issuecomment-504557094

app/views/blocks/campusQuestion.html
```diff
@@ -2,14 +2,18 @@
...
     ng-required="block.required"
     uib-typeahead="campus for campus in searchCampuses($viewValue)"
     typeahead-loading="loadingCampuses"
-    typeahead-min-length="4"
-    typeahead-wait-ms="300"
+    typeahead-min-length="2"
+    typeahead-wait-ms="100"
+    typeahead-editable="false"
+    typeahead-no-results="campusNotFound"
+    typeahead-show-hint="true"
```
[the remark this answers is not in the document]
caleballdrin: Just added that

app/views/blocks/campusQuestion.html
```diff
@@ -2,14 +2,18 @@
...
     placeholder=""
     class="form-control"
+    show-errors
     ng-required="block.required"
     uib-typeahead="campus for campus in searchCampuses($viewValue)"
     typeahead-loading="loadingCampuses"
-    typeahead-min-length="4"
-    typeahead-wait-ms="300"
+    typeahead-min-length="2"
+    typeahead-wait-ms="100"
```
knutsenm, line 13: Does that query ERT (and in turn, Infobase) 10x/sec? Ick.

wrandall22: The typeahead wait time is how long between the time they type a character and when the query to ERT's API happens. It still only does one query per character max, in this case it would be upwards of 10x/sec if they are typing that much at that pace.

caleballdrin: @wrandall22 Would it be better to load all the campus names into an array when the form loads?

dr-bizz: I don't think downloading 5,000 or 25,000 campuses is the right direction to go.

You mention, `The main thing is that we want the user to see that it is an autocomplete box when they start typing.`

We can do a few things to achieve that:
- Add small text under the input stating it's an autocomplete box.
- On the initial input focus, show the autocomplete box with a list of the most popular schools people select.

hlbraddock: I'm fine with @dr-bizz proposal.

test/spec/directives/block.spec.js
```diff
@@ -157,4 +157,36 @@ describe('Directive: blocks', () => {
...
+      $compile('<campus-question></campus-question>')($scope);
+      $scope.$digest();
+
+      $scope.searchCampuses('San');
+
+      expect($scope.params.limit).toBeDefined();
+
+      expect($scope.params.includeInternational).toBeDefined();
+    });
+  });
```
dr-bizz, line 191: Should test without showInternationalCampuses as well.
```suggestion
    it('searchCampuses params are formed correctly', () => {
      $scope.block.content.showInternationalCampuses = true;
      $compile('<campus-question></campus-question>')($scope);
      $scope.$digest();

      $scope.searchCampuses('San');

      expect($scope.params.limit).toBeDefined();

      expect($scope.params.includeInternational).toBeDefined();
    });

    it("doesn't add includeInternational", () => {
      $compile('<campus-question></campus-question>')($scope);
      $scope.$digest();

      $scope.searchCampuses('San');

      expect($scope.params.includeInternational).not.toBeDefined();
    });
```

app/scripts/directives/questionToolbar.js
```diff
@@ -170,16 +170,18 @@ angular
           },
           {
             id: 'campusQuestion',
-            defaultTitle: 'Campus',
+            defaultTitle:
+              'Campus- Type the full name of your campus and choose from the list that appears',
```
dr-bizz, line 174: I believe we need a space between `Campus` and the dash.
```suggestion
              'Campus - Type the full name of your campus and choose from the list that appears',
```

app/scripts/directives/questionToolbar.js
```diff
@@ -170,16 +170,18 @@ angular
...
             defaultProfile: 'CAMPUS',
-            iconClass: 'fa-graduation-cap',
+            iconClass: 'fa-university',
             name: 'Campus',
+            defaultExportFieldTitle: 'Campus',
```
dr-bizz, line 178: This is more of a question. What is the need of `defaultExportFieldTitle`, it doesn't seem to do anything.

app/views/blocks/campusQuestion.html
caleballdrin: @dr-bizz I'm not sure if we'd want to show popular schools. Right now it shows an error message when they type in something not in the list. So that may be enough to get them to realize it is an autocomplete box without needing to explicitly say that.

I will change it to 200.

dr-bizz (changes requested):

app/scripts/directives/questionToolbar.js
caleballdrin: Since the default title for the question is so long, the product owner requested to automatically change the title to something shorter when it is exported.

test/spec/directives/block.spec.js
caleballdrin: Ok I'll add that. thanks

test/spec/directives/block.spec.js
```diff
@@ -157,4 +157,36 @@ describe('Directive: blocks', () => {
...
+
+      $scope = $rootScope.$new();
+      $templateCache.put('views/blocks/campusQuestion.html', '');
+      $scope.block = _.cloneDeep(
+        testData.conference.registrationPages[1].blocks[4],
+      );
+      $scope.block.content.showInternationalCampuses = true;
+    }));
+
+    it('searchCampuses params are formed correctly', () => {
```
wrandall22, line 181:
```suggestion
    it('forms the searchCampuses params correctly', () => {
```

dr-bizz (approved):
Great work!

PR #838, opened 2023-08-08
EVENT-826 New Profile Questions

caleballdrin:
## Description
Adding 3 new profile questions:

**Expected Graduation Date**
https://jira.cru.org/browse/EVENT-826
- New question that can be added to the top of the Questions page "Graduation Date"
- The quesiton can then be dragged into place and tagged as a Cru Profile Question.
- Question Default Title: "Expected Graduation Year (month & year)"
- Question should be a drop down for month/year.

**Ethnicity**
https://jira.cru.org/browse/EVENT-827
- New question that can be added to the top of the Questions page "Racial/Ethnic Identity"
- The quesiton can then be dragged into place and tagged as a Cru Profile Question.
- Question Default Title: "Racial/Ethnic Identity"
- Question can be considered a "Checkbox" "Multiple Choice" question so that multiple options one option can be selected from the options
- The custom "Other" Option does need to be on by default in the ERT but if it could be added for this default question that would be desired.

**Cru Opportunities**
https://jira.cru.org/browse/EVENT-828
- New question that can be added to the top of the Questions page "Cru Opportunities"
- The quesiton can then be dragged into place and tagged as a Cru Profile Question.
- Question Default Title: "Would you like to be informed about new opportunities (like conferences, mission trips, career opportunities, etc.) to be involved with Cru around the U.S. & world? Checking yes below will add you to a subscription list to receive monthly updates on the latest Cru opportunities! You can unsubscribe at any time."
- Question can be considered a "Multiple Choice" quesiton and expect one answer

## Changes I've made
- Added 3 new profile questions
- Modified date picker to show the graduation date picker in a different format
- Copied the radioQuestion to make ethnicityQuestion
- Add the 3 profile questions as possible rules

### DIFF
--- a/app/scripts/directives/blocks.js
+++ b/app/scripts/directives/blocks.js
@@ -270,32 +270,6 @@ angular
     return {
       templateUrl: graduationDateQuestionTemplate,
       restrict: 'E',
-      controller: function ($scope) {
-        $scope.answer.value = $scope.answer.value
-          ? Date.parse($scope.answer.value, 'yyyy-MM-dd')
-          : '';
-
-        $scope.dateOptions = {
-          viewMode: 'years',
-          datepickerMode: 'year',
-          minMode: 'month',
-          yearRows: 3,
-          yearColumns: 4,
-        };
-
-        $scope.altInputFormats = [
-          'dd-MMMM-yyyy',
-          'yyyy/MM/dd',
-          'dd.MM.yyyy',
-          'shortDate',
-          'yyyy-mm-dd',
-          'MMMM yyyy',
-        ];
-
-        $scope.open = function () {
-          $scope.opened = true;
-        };
-      },
     };
   });
 
--- a/app/scripts/directives/blocks.js
+++ b/app/scripts/directives/blocks.js
@@ -253,6 +253,32 @@ angular
     return {
       templateUrl: graduationDateQuestionTemplate,
       restrict: 'E',
+      controller: function ($scope) {
+        $scope.answer.value = $scope.answer.value
+          ? Date.parse($scope.answer.value, 'yyyy-MM-dd')
+          : '';
+
+        $scope.dateOptions = {
+          viewMode: 'years',
+          datepickerMode: 'year',
+          minMode: 'month',
+          yearRows: 3,
+          yearColumns: 4,
+        };
+
+        $scope.altInputFormats = [
+          'dd-MMMM-yyyy',
+          'yyyy/MM/dd',
+          'dd.MM.yyyy',
+          'shortDate',
+          'yyyy-mm-dd',
+          'MMMM yyyy',
+        ];
+
+        $scope.open = function () {
+          $scope.opened = true;
+        };
+      },
     };
   });
 
--- a/app/scripts/directives/blocks.js
+++ b/app/scripts/directives/blocks.js
@@ -299,8 +299,9 @@ angular
           }
         };
 
-        $scope.clearAnswer = () => {
-          $scope.selectedAnswer = '';
+        $scope.clearOther = () => {
+          if ($scope.selectedAnswer != $scope.otherSentinel)
+            $scope.otherAnswer = '';
         };
 
         const answerValue = $scope.answer ? $scope.answer.value : '';
--- a/app/scripts/directives/blocks.js
+++ b/app/scripts/directives/blocks.js
@@ -13,6 +13,7 @@ import yearInSchoolQuestionTemplate from 'views/blocks/yearInSchoolQuestion.html
 import opportunitiesQuestionTemplate from 'views/blocks/opportunitiesQuestion.html';
 import textareaQuestionTemplate from 'views/blocks/textareaQuestion.html';
 import campusQuestionTemplate from 'views/blocks/campusQuestion.html';
+import graduationDateQuestionTemplate from 'views/blocks/graduationDateQuestion.html';
 
 angular.module('confRegistrationWebApp').directive('nameQuestion', function () {
   return {
@@ -241,6 +242,15 @@ angular
   .directive('opportunitiesQuestion', function () {
     return {
       templateUrl: opportunitiesQuestionTemplate,
+      restrict: 'E',
+    };
+  });
+
+angular
+  .module('confRegistrationWebApp')
+  .directive('graduationDateQuestion', function () {
+    return {
+      templateUrl: graduationDateQuestionTemplate,
       restrict: 'E',
     };
   });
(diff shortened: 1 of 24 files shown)


app/scripts/directives/datepicker.js
```diff
@@ -10,25 +10,51 @@ angular
...
+                'MM/YYYY',
+                'MM-YY',
+                'MM-YYYY',
+                'MMM-YYYY',
+                'MMMM-YYYY',
+              ],
+            }
+          : {
+              defaultDate: initialDate,
+            };
```
dr-bizz, line 49: You could rename graduationDate to `dateFormat` and use it in the code as the format? Just a thought.
```suggestion
          scope.localModel ? moment(new Date(scope.localModel)).format(`${scope.dateFormat} hh:mm A`)
            : null;
        let dateOptions = scope.dateFormat === 'MMMM-YYYY'
          ? {
              viewMode: 'years',
              format: dateFormat,
              defaultDate: initialDate,
              useCurrent: false,
              extraFormats: [
                'MM/YY',
                'MM/YYYY',
                'MM-YY',
                'MM-YYYY',
                'MMM-YYYY',
                'MMMM-YYYY',
              ],
            }
          : {
              defaultDate: initialDate,
            };
```

app/scripts/directives/rule.js
```diff
@@ -129,15 +132,13 @@ angular.module('confRegistrationWebApp').directive('rule', function () {
...
             return ['M', 'F'];
           case 'yearInSchoolQuestion':
-            return [
-              'Freshman',
-              'Sophomore',
-              'Junior',
-              'Senior',
-              'Graduate Student',
-            ];
+            return block.content.staticChoices;
```
dr-bizz, line 135: Currently, the Graduation Date is showing no Question rules, is that intentional?

app/scripts/directives/blocks.js
```diff
@@ -234,3 +246,104 @@ angular
...
+  .module('confRegistrationWebApp')
+  .directive('ethnicityQuestion', function () {
+    return {
+      templateUrl: ethnicityQuestionTemplate,
+      restrict: 'E',
+      controller: function ($scope, $timeout) {
+        $scope.otherSentinel = '__other__';
+
+        $scope.block.content.staticChoices = [
+          'American Indian/Native Alaskan',
```
dr-bizz, line 286: I think they prefer `Native American` but I would double check.

app/scripts/directives/blocks.js
```diff
@@ -234,3 +246,104 @@ angular
...
+        $scope.clearOther = () => {
+          if ($scope.selectedAnswer != $scope.otherSentinel)
+            $scope.otherAnswer = '';
+        };
+
+        const answerValue = $scope.answer ? $scope.answer.value : '';
+        if (
+          answerValue &&
+          !_.includes($scope.block.content.staticChoices, answerValue)
+        ) {
```
dr-bizz, line 311: We need to make an other choice, which will make this IF statement easier to understand.

app/scripts/directives/blocks.js
```diff
@@ -234,3 +246,104 @@ angular
...
+          // Update the answer after one second
+          currentTimeout = $timeout(() => {
+            if ($scope.answer) {
+              $scope.answer.value =
+                $scope.selectedAnswer === $scope.otherSentinel
+                  ? $scope.otherAnswer
+                  : $scope.selectedAnswer;
+            }
+          }, 300);
+        };
```
dr-bizz, line 341: I believe we have the timeout functionailty in multiple places. it might be worth creating a Lib file which handles the timeout in which you pass a callback which runs once the timout has completed.

app/scripts/directives/blocks.js
dr-bizz: You will be able to make the function `selectOtherAnswer` simpler as well.

app/views/blocks/graduationDateQuestion.html
```diff
@@ -0,0 +1,8 @@
+<div class="row">
+  <div class="col-sm-4 stacked-spacing-col-sm">
+    <crs-datetimepicker
+      model="answer.value"
+      graduation-date="true"
```
dr-bizz, line 5: As this is very different to all the other dates. I wonder if we need to make the date modal appear on focus as the user might be confused about what to type.

app/views/blocks/opportunitiesQuestion.html
```diff
@@ -0,0 +1,18 @@
...
+    <input
+      type="radio"
+      show-errors
+      name="{{ block.id }}"
+      ng-required="block.required"
+      ng-model="answer.value"
+      ng-model-options="{ debounce: 300 }"
+      ng-value="choice"
+    />
+    <strong>{{ choice }}</strong></label
```
dr-bizz, line 16: Should choosing `Yes, via text` and `Yes, via email and text` open a input for the user to enter their telephone?

dr-bizz (changes requested):
The code is looking good. I have a few comments.

app/scripts/directives/blocks.js
caleballdrin: This was the official list they got from HR.

app/scripts/directives/rule.js
caleballdrin: Graduation date should have question rules. The question has to be below the graduation date question. It is working for me.

app/views/blocks/opportunitiesQuestion.html
caleballdrin: They can add a Phone number profile question to get than info the same way they get their email.

app/scripts/directives/blocks.js
caleballdrin: I'm not sure how to do that. Could you show an example?

app/scripts/directives/datepicker.js
caleballdrin: I think that could be a little confusing because people might think that setting dateFormat could change the format to whatever they want. When there is actually only 1 value they can provide.

app/scripts/directives/rule.js
caleballdrin: Questions related to dates aren't in this switch. This is only for questions that have choices to choose from.

app/views/blocks/ethnicityQuestion.html
```diff
@@ -0,0 +1,42 @@
...
+    class="form-control"
+    show-errors
+    ng-model="otherAnswer"
+    ng-change="selectOtherAnswer()"
+    ng-required="selectedAnswer === otherSentinel"
+    style="display: inline; width: auto"
+    autocomplete="off"
+    aria-autocomplete="none"
+  />
+</label>
```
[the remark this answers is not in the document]
caleballdrin: I added this change

app/scripts/directives/blocks.js
dr-bizz: ah okay use that then

dr-bizz (approved):
Great work on this!

dr-bizz (approved):

PR #839, opened 2023-08-14
A recent deployment broke some code on the event details page, result…

dr-bizz:
## Description
A recent deployment broke some code on the event details page, resulting in the Associated Registrant Types no longer showing. This fix adds back the code which was removed.

## Changes
- Adding back code which was there previously. Ensuring `name` and `selected` are defined on `types` in the  `allowedRegistrantTypeSet` array.

Deployment which had breaking changes https://github.com/CruGlobal/conf-registration-web/pull/835

### DIFF
--- a/app/scripts/controllers/eventDetails.js
+++ b/app/scripts/controllers/eventDetails.js
@@ -913,13 +913,14 @@ angular
         });
         const payload = angular.copy($scope.conference);
         // Remove unwanted properties from sending to API.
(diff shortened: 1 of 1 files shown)


wrandall22:
Does this break the liability form again, which is why Caleb removed it?

dr-bizz:
Looks like there is an error

dr-bizz:
Fixed the error which Caleb described.

wrandall22 (approved):

PR #841, opened 2023-08-29
Fix Date Format

caleballdrin:
Change the format of saved data for the graduation date question. The API wants YYYY-MM as opposed to the usual date format that includes the day and time.

### DIFF
--- a/app/scripts/directives/datepicker.js
+++ b/app/scripts/directives/datepicker.js
@@ -15,8 +15,11 @@ angular
       controller: function ($timeout, $scope) {
         $scope.updateTimeStamp = function (timestamp) {
           $scope.$apply(function () {
+            let dateSaveFormat = $scope.monthYearOnly
+              ? 'YYYY-MM'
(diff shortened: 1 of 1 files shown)


caleballdrin:
@dr-bizz can you help give me ideas for how to get this test to work? I can't seem to be able to access any of the variables or functions like $scope.updateTimeStamp('2020-01-01');

dr-bizz:
With my latest push, you're able to get `updateTimeStamp` and the other values on `scope`. Still needs some work doing, but I hope this has sent you in the right direction.

caleballdrin:
@dr-bizz thanks for doing that. This should be ready for review. I talked to Lee and he would like the format to include the day for the API. So I had to hide the day when the answer is displayed.

caleballdrin:
@dr-bizz This works as it should on my local but it is not working on staging. I wonder if staging didn't update for some reason.

caleballdrin:
@dr-bizz it is now on staging.

dr-bizz (approved):

PR #842, opened 2023-09-07
Fix Graduation date question and email question

caleballdrin:
## Description
- Fix the bug in https://secure.helpscout.net/conversation/2353023844/1012529/
- Fix graduation date format display

## Changes I've made
- On initialization of the graduation date question, format the value that we get from the API

### DIFF
--- a/app/scripts/directives/datepicker.js
+++ b/app/scripts/directives/datepicker.js
@@ -16,7 +16,7 @@ angular
         $scope.updateTimeStamp = function (timestamp) {
           //For Graduation date question, set the day to 1. The API needs the day but that could change in the future.
           timestamp = $scope.monthYearOnly
-            ? moment(new Date(timestamp)).set('date', 1)
+            ? moment(new Date(timestamp)).set('date', 10)
             : timestamp;
           $scope.$apply(function () {
             let dateSaveFormat = $scope.monthYearOnly
@@ -30,12 +30,13 @@ angular
       },
       link: function (scope, element) {
         var datePickerElement = angular.element(element).find('.datepicker');
-        var initialDate =
+        scope.localModel =
           scope.localModel && scope.monthYearOnly
-            ? scope.localModel
+            ? moment(new Date(scope.localModel)).format('YYYY-MM-DD')
             : scope.localModel
             ? moment(new Date(scope.localModel)).format('MM/DD/YYYY hh:mm A')
             : null;
+        var initialDate = scope.localModel ? scope.localModel : null;
         scope.dateOptions = scope.monthYearOnly
           ? {
               viewMode: 'years',
--- a/test/spec/directives/datepicker.spec.js
+++ b/test/spec/directives/datepicker.spec.js
@@ -21,7 +21,7 @@ describe('Directive: datepicker', function () {
   it('Sets the date to the correct format based on the type of date question', function () {
     scope.updateTimeStamp(new Date('02/05/1994'));
 
-    expect(scope.localModel).toBe('1994-02-01');
+    expect(scope.localModel).toBe('1994-02-10');
 
     scope.monthYearOnly = false;
 


app/scripts/directives/datepicker.js
```diff
@@ -16,7 +16,7 @@ angular
         $scope.updateTimeStamp = function (timestamp) {
           //For Graduation date question, set the day to 1. The API needs the day but that could change in the future.
           timestamp = $scope.monthYearOnly
-            ? moment(new Date(timestamp)).set('date', 1)
+            ? moment(new Date(timestamp)).set('date', 10)
```
wrandall22, line 19: If this is changing, the comment above should also change.

caleballdrin: Ah yes good catch

app/scripts/directives/datepicker.js
```diff
@@ -30,12 +30,13 @@ angular
...
         var datePickerElement = angular.element(element).find('.datepicker');
-        var initialDate =
+        scope.localModel =
           scope.localModel && scope.monthYearOnly
-            ? scope.localModel
+            ? moment(new Date(scope.localModel)).format('YYYY-MM-DD')
             : scope.localModel
             ? moment(new Date(scope.localModel)).format('MM/DD/YYYY hh:mm A')
             : null;
+        var initialDate = scope.localModel ? scope.localModel : null;
```
wrandall22, line 39: Should this really be a `var`?

wrandall22 (commented):
Not sure I understand how using the 10th instead of the 1st helps fix the problem. Is this due to time zone issues? Is the 10th an arbitrary day of the month that is not used for anything but required by the API?

caleballdrin:
@wrandall22 It doesn't matter what the day is, the API just needs a day. The time and timezone are set to 0 by the API. When I bring the data back in from the API and format it again, the datepicker sets it to the previous month. So I changed it to 10.

app/scripts/directives/datepicker.js
caleballdrin: Just changed it

app/scripts/directives/datepicker.js
```diff
@@ -36,7 +36,7 @@ angular
             : scope.localModel
             ? moment(new Date(scope.localModel)).format('MM/DD/YYYY hh:mm A')
             : null;
-        var initialDate = scope.localModel ? scope.localModel : null;
+        let initialDate = scope.localModel ? scope.localModel : null;
```
wrandall22, line 39: In an ideal world, this change would have happened in a separate commit, then squashed into the prior one where the change for this happened in order to keep clean git history. Not worth the effort to change it now.

caleballdrin: Oh ok. Would I have needed to squash locally and then force push the changed commit?

app/scripts/controllers/registration.js
```diff
@@ -375,9 +375,16 @@ angular
             var savedAnswer = _.find(currentRegistrantOriginalAnswers, {
               id: a.id,
             });
+            let invalidBlocks = validateRegistrant.validate(
```
wrandall22, line 378: Looks like this can be `const`.

app/scripts/controllers/registration.js
```diff
@@ -375,9 +375,16 @@ angular
...
+              _.find(currentRegistration.registrants, {
+                id: currentRegistrant.id,
+              }),
+            );
             if (
-              angular.isUndefined(savedAnswer) ||
-              !angular.equals(savedAnswer.value, a.value)
+              (angular.isUndefined(savedAnswer) ||
+                !angular.equals(savedAnswer.value, a.value)) &&
+              !invalidBlocks.includes(a.blockId)
```
wrandall22, line 387: Is there a chance `invalidBlocks` could come back null or undefined?

caleballdrin: Hmm I don't think so. But I will check to make sure it finds a registrant before assigning it.

app/scripts/directives/datepicker.js
```diff
@@ -30,12 +30,13 @@ angular
...
         var datePickerElement = angular.element(element).find('.datepicker');
-        var initialDate =
+        scope.localModel =
           scope.localModel && scope.monthYearOnly
-            ? scope.localModel
+            ? moment(new Date(scope.localModel)).format('YYYY-MM-DD')
             : scope.localModel
             ? moment(new Date(scope.localModel)).format('MM/DD/YYYY hh:mm A')
             : null;
+        let initialDate = scope.localModel ? scope.localModel : null;
```
wrandall22, line 39: I'm not the best at concise JS syntax, but would this work?
```
const initialDate = scope.localModal || null;
```

caleballdrin: yes i believe so

wrandall22 (approved):

PR #843, opened 2023-09-08
Revert "Fix Graduation date question and saving questions validation (#842)"

dr-bizz:
This commit causes all event details to an error on a date field when the user ties to save.

This reverts commit 0db31d84fb88f83057af1544067c3fafb22d046b.

### DIFF
--- a/app/scripts/controllers/registration.js
+++ b/app/scripts/controllers/registration.js
@@ -367,16 +367,6 @@ angular
         var currentRegistrant = _.find($scope.currentRegistration.registrants, {
           id: $scope.currentRegistrant,
         });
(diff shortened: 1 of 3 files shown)


wrandall22:
Hmm, we need to figure out a solution to the timezone issue Caleb was attempting to fix as well.

wrandall22 (approved):

dr-bizz:
It seems `app/scripts/directives/datepicker.js` is the file which is causing the error

wrandall22:
We should probably boost our automated testing to catch this kind of error in the future.

dr-bizz:
Yes, we need to beef up ERT's tests and Synthetic Tests to ensure mistakes like these are caught.

PR #844, opened 2023-09-08
Reimplementing the reverted PR #842/#843 but not included changes wer…

dr-bizz:
## Description
Re-adding changes which were on PR #842 but not included changes where localModel is set.
https://secure.helpscout.net/conversation/2355398292/1014142?folderId=7296147

## Changes
- Check to see if questions are invalid before pushing them to the database.
- Set graduation date day to 10 so it's the same month across all timezones.

### DIFF
--- a/app/scripts/controllers/registration.js
+++ b/app/scripts/controllers/registration.js
@@ -367,6 +367,16 @@ angular
         var currentRegistrant = _.find($scope.currentRegistration.registrants, {
           id: $scope.currentRegistrant,
         });
+        const invalidBlocks = _.find(currentRegistration.registrants, {
+          id: currentRegistrant.id,
+        })
+          ? validateRegistrant.validate(
+              conference,
+              _.find(currentRegistration.registrants, {
+                id: currentRegistrant.id,
+              }),
+            )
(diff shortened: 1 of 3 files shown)


wrandall22:
Were you able to reproduce the issue from Helpscout before changes then confirm this does not have similar issues?

dr-bizz:
Yes, the bug is in the code below, as Caleb was updating `scope.localModel` rather than just setting the initial value.

```
scope.localModel =
          scope.localModel && scope.monthYearOnly
            ? moment(new Date(scope.localModel)).format('YYYY-MM-DD')
            : scope.localModel
            ? moment(new Date(scope.localModel)).format('MM/DD/YYYY hh:mm A')
            : null;
let initialDate = scope.localModel || null;
```

https://github.com/CruGlobal/conf-registration-web/pull/842/files#diff-5ba14cf6c1e9ade6d0e7985ce3c5047d7e2069d7576426304725b7b4ceb13719R33-R39

wrandall22 (approved):

dr-bizz:
Tested on staging and looks good to go.

PR #845, opened 2023-09-12
Date format fix

caleballdrin:
## Description
The Graduation date question displays the entire date/timestamp from the API when looking at past registrations. When displaying the dates, format them to make them more readable.

## Changes made
- Removed logic around formatting the initial date when loading a date picker. The full date/timestamp is fine. Each datepicker formats the displayed date based on what kind of question it is.
- Formatted the dates in a readable format when they are displayed throughout the ERT.
- Added new tests to check the initialDate

### DIFF
--- a/app/scripts/directives/datepicker.js
+++ b/app/scripts/directives/datepicker.js
@@ -27,9 +27,9 @@ angular
             );
           });
         };
-        $scope.getDateOptions = function (model) {
-          let initialDate = model
-            ? moment(model).format('YYYY-MM-DD hh:mm:ss')
+        $scope.getDateOptions = function () {
+          let initialDate = $scope.localModel
+            ? moment($scope.localModel).format('YYYY-MM-DD HH:mm:ss')
             : null;
           const dateOptions = $scope.monthYearOnly
             ? {
--- a/app/scripts/directives/datepicker.js
+++ b/app/scripts/directives/datepicker.js
@@ -30,12 +30,9 @@ angular
       },
       link: function (scope, element) {
         var datePickerElement = angular.element(element).find('.datepicker');
-        var initialDate =
-          scope.localModel && scope.monthYearOnly
-            ? scope.localModel
-            : scope.localModel
-            ? moment(new Date(scope.localModel)).format('MM/DD/YYYY hh:mm A')
-            : null;
+        var initialDate = scope.localModel
+          ? moment(new Date(scope.localModel)).format('YYYY-MM-DD HH:mm:ss')
+          : null;
         scope.dateOptions = scope.monthYearOnly
           ? {
               viewMode: 'years',
--- a/app/scripts/directives/datepicker.js
+++ b/app/scripts/directives/datepicker.js
@@ -27,32 +27,36 @@ angular
             );
           });
         };
+        $scope.getDateOptions = function (model) {
+          let initialDate = model
+            ? moment(model).format('YYYY-MM-DD hh:mm:ss')
+            : null;
+          const dateOptions = $scope.monthYearOnly
+            ? {
+                viewMode: 'years',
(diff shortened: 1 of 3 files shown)


app/scripts/directives/datepicker.js
```diff
@@ -30,12 +30,9 @@ angular
...
         var datePickerElement = angular.element(element).find('.datepicker');
-        var initialDate =
-          scope.localModel && scope.monthYearOnly
-            ? scope.localModel
-            : scope.localModel
-            ? moment(new Date(scope.localModel)).format('MM/DD/YYYY hh:mm A')
-            : null;
+        var initialDate = scope.localModel
+          ? moment(new Date(scope.localModel)).format('YYYY-MM-DD HH:mm:ss')
+          : null;
```
dr-bizz, line 35: If you're setting the variable `initialDate` back to what it was previously, below is what it was. The `A` just stands for AM/PM, but the inputs show it anyway, even when you remove it. (I haven't checked all inputs but most)

```suggestion
        var initialDate = scope.localModel
          ? moment(scope.localModel).format('YYYY-MM-DD HH:mm A')
          : null;
```

dr-bizz (changes requested):
We need a write multiple tests and maybe multiple test pages for this change since it affects a lot of functionality.

app/scripts/directives/datepicker.js
caleballdrin: Setting the initial date to a certain format is actually just to make sure the datepicker can read it. The datepicker itself sets the format. So each datepicker in different parts of the page will format the display of the date in the datepicker differently.

caleballdrin:
@dr-bizz I reorganized some of the tests to make them more clear. Are these enough tests? What other things should I test?

app/scripts/directives/datepicker.js
```diff
@@ -27,35 +27,36 @@ angular
             );
           });
         };
+        $scope.getDateOptions = function (model) {
+          let initialDate = model
```
dr-bizz, line 31: Are you able to do `$scope.localModel` here, or do we need to pass in modal?

test/spec/directives/datepicker.spec.js
```diff
@@ -18,19 +18,33 @@ describe('Directive: datepicker', function () {
...
     expect(scope.localModel).toBe('1994-02-05 00:00:00');
   });
 
-  it('Sets date options correctly based on type of date', function () {
-    expect(scope.dateOptions.viewMode).toBe('years');
+  it('Sets the initial/default date to the correct value and format based on the localModel', function () {
+    let localModel = new Date('01-30-2023 11:01:05');
+    let dateOptions = scope.getDateOptions(localModel);
+
+    expect(dateOptions.defaultDate).toBe('2023-01-30 11:01:05');
```
dr-bizz, line 48: Sorry to be a stickler about this, but can you add a few more tests to ensure `getDateOptions` is acting correctly?

- Ensure the whole object comes back the way you intend.
- If the API returns month & year dates differently, a test with a date formatted like that.
- Test without month & year to ensure getDateOptions returns correctly.

dr-bizz (changes requested):

test/spec/directives/datepicker.spec.js
caleballdrin: done

app/scripts/directives/datepicker.js
caleballdrin: i just changed that

dr-bizz (commented):
This looks good to me. Can you push this to the stage and let me know when you have so we can test it there before it goes live?

caleballdrin:
@dr-bizz i put it on staging yesterday. So you can test whenever.

dr-bizz (approved):

PR #847, opened 2023-11-06
No-Jira Fix Campus

caleballdrin:
## Description
Students are submitting ERT forms with campuses that do not match the campus names in Infobase. This is because some browsers are auto-filling the campus data. The UI only validates when the user changes the field. So this bad data gets submitted.

## Changes I made
- On page load, check if there is an answer already in the field. If there is, check to make sure that exists in the database with the searchCampuses() function. If it doesn't, set the answer to ''.
- Every time the UI validates the fields, include the campus question in that validation to make sure it is not empty.


app/scripts/directives/blocks.js
```diff
@@ -251,6 +251,13 @@ angular
               return campusNames.data;
             });
         };
+        if ($scope.answer.value) {
+          $scope.searchCampuses($scope.answer.value).then((data) => {
+            if (data.length < 1) {
```
canac, line 256: Do you think this conditional could be clearer as `data.length === 0`?

caleballdrin: yes I'll change that

canac (changes requested):
One stray comment, but it looks great! Good work with the tests too.

canac:
Unless it's on purpose, you probably want to revert all your changes to `package.json` and `yarn.lock`. I believe that `git checkout master -- package.json yarn.lock` should do it.

app/scripts/directives/blocks.js
```diff
@@ -251,6 +251,14 @@ angular
               return campusNames.data;
             });
         };
+        if ($scope.answer.value) {
+          $scope.searchingCampuses = true;
+          $scope.searchCampuses($scope.answer.value).then((data) => {
+            if (data.length === 0) {
```
dr-bizz, line 257: In JS, `0` is `false`, so if the length is 0 run line 258, if 1 or above skip it.
```suggestion
            if (!data.length) {
```

dr-bizz (changes requested):
When testing on staging, my autofill still works. I was able to get to the final stage of the register form.

Screenshot 2023-11-08 at 8 31 21 AM

caleballdrin:
@dr-bizz I haven't pushed it to staging yet. The On Staging label doesn't work for the ERT. I'm not sure why. It would be nice if it did.

caleballdrin:
@dr-bizz It is on staging now.

app/scripts/directives/blocks.js
```diff
@@ -251,6 +251,14 @@ angular
               return campusNames.data;
             });
         };
+        if ($scope.answer.value) {
+          $scope.searchingCampuses = true;
```
canac, line 255: Setting a variable just so that it can be checked in a test isn't a pattern I'm used to seeing. Can you remove that line and update the test something like the one I wrote in the other comment?

dr-bizz:
Just tested it again and this is still happening

dr-bizz (changes requested):
Just tested it again and this is still happening

test/spec/directives/block.spec.js
```diff
@@ -199,6 +199,16 @@ describe('Directive: blocks', () => {
 
       expect($scope.params.includeInternational).not.toBeDefined();
     });
+
+    it('checks the campus database if a answer is present on page load', () => {
```
canac, line 203: I think that a test like this would be better. It mocks the network call to return `[]` (i.e. no campuses) and checks that the answer was set back to blank.

```js
    it('checks the campus database if a answer is present on page load', () => {
      $httpBackend
        .whenGET('campuses/SFSU?includeInternational=true&limit=15')
        .respond(() => [200, []]);

      $scope.answer = { value: 'SFSU' };
      $compile('<campus-question></campus-question>')($scope);

      $httpBackend.flush();
      $httpBackend.verifyNoOutstandingExpectation();
      $httpBackend.verifyNoOutstandingRequest();

      expect($scope.answer.value).toBe('');
    });
```

The `forms the searchCampuses params correctly` and `doesn't add includeInternational` tests could also use network mocking instead of checking $scope.params, but that doesn't need to be done in this PR.

canac (changes requested):
There's one more tweak we should make to the test. I think that if you use my suggestion, lines 257-258 will be hit by code coverage. We also might need a test with a non-empty campus answer in `validateRegistrant.spec.js` to cover line 343.

canac (approved):
Looks good now! 💯

dr-bizz (approved):

PR #849, opened 2023-12-12
EVENT-843 - FedCM Beta Testing

caleballdrin:
## Description
https://jira.cru.org/browse/EVENT-843
Google login is changing and we need to be ready for it, we should hop on the FedCM beta to make sure this move won't break ERT's ability to login via Google.

https://developers.googleblog.com/2023/08/announcing-federated-credential-management-beta-for-gis.html

## Changes I made
I followed these steps: https://developers.google.com/identity/gsi/web/guides/fedcm-migration

### DIFF
--- a/app/.htaccess
+++ b/app/.htaccess
@@ -278,7 +278,7 @@ AddDefaultCharset utf-8
 # see the specification: http://w3.org/TR/CSP).
 
 # <IfModule mod_headers.c>
-#    Header set Content-Security-Policy "script-src 'self'; object-src 'self'"
+#    Header set Content-Security-Policy "script-src 'self' https://accounts.google.com/gsi/client; object-src 'self'"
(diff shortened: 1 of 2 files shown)


app/.htaccess
```diff
@@ -278,7 +278,7 @@ AddDefaultCharset utf-8
 # see the specification: http://w3.org/TR/CSP).
 
 # <IfModule mod_headers.c>
-#    Header set Content-Security-Policy "script-src 'self'; object-src 'self'"
+#    Header set Content-Security-Policy "script-src 'self' https://accounts.google.com/gsi/client; object-src 'self'"
```
canac, line 281: Does this change do anything? Isn't this entire line just a comment?

caleballdrin: I guess it doesn't do anything. But if the CSP ever gets used, it will need to be there.

canac (approved):
I was able to login with Google in a Chromium browser (which I believe supports FedCM) and in Firefox (which I don't believe supports FedCM).

wrandall22:
[Compatibility reference](https://developer.mozilla.org/en-US/docs/Web/API/FedCM_API#browser_compatibility)

wrandall22 (approved):

PR #834, opened 2023-06-19
Updating cru payments

dr-bizz:
I am updating the NPM package `cru-payments` to the new updated security patched version `@CruGlobal/cru-payments`.

- Removed old package `cru-payments`
- Added new package `@CruGlobal/cru-payments`
- Configured new package.
- **Added logs while testing on staging. Will remove it before deployment.**

PR #850, opened 2023-12-19
[no-Jira] Sync yarn.lock and package.json

canac:
It looks like yarn.lock and package.json got out of sync after #834. Every time I run `yarn install`, yarn makes these changes to the lockfile.


dr-bizz (approved):

dr-bizz:
Thank you for doing this

PR #851, opened 2023-12-20
[no-Jira] Improve test coverage

canac:
Increase the test coverage of some of the most frequently-modified files.

### DIFF
--- a/app/scripts/controllers/eventForm.js
+++ b/app/scripts/controllers/eventForm.js
@@ -264,6 +264,7 @@ angular
 
         var profileType = null;
         if (angular.isDefined(defaultProfile)) {
+          // Check whether a block with this profile type exists already on any page
           var profileCount = 0;
           $scope.conference.registrationPages.forEach(function (page) {
             page.blocks.forEach(function (block) {
--- a/app/scripts/controllers/eventForm.js
+++ b/app/scripts/controllers/eventForm.js
@@ -5,7 +5,6 @@ angular
     function (
       $rootScope,
       $scope,
-      $uibModal,
       modalMessage,
       $location,
       $anchorScroll,
@@ -46,14 +45,13 @@ angular
       var formSavingNotifyTimeout;
 
       function saveForm() {
+        $timeout.cancel(formSavingTimeout);
         if (formSaving) {
-          $timeout.cancel(formSavingTimeout);
           formSavingTimeout = $timeout(function () {
             saveForm();
           }, 600);
           return;
         }
-        $timeout.cancel(formSavingTimeout);
 
         formSaving = true;
         let conferenceWithoutImage = angular.copy($scope.conference);
@@ -210,28 +208,29 @@ angular
           });
       };
 
-      var makePositionArray = function () {
-        var tempPositionArray = [];
+      // Generate a map of blockIds to their page and index within the page
+      var makeBlockPositionMap = function () {
+        var positionMap = {};
         $scope.conference.registrationPages.forEach(function (page, pageIndex) {
           page.blocks.forEach(function (block, blockIndex) {
-            tempPositionArray[block.id] = {
-              page: pageIndex,
-              block: blockIndex,
+            positionMap[block.id] = {
+              pageIndex,
+              blockIndex,
             };
           });
         });
-        return tempPositionArray;
+        return positionMap;
       };
 
       $scope.copyBlock = function (blockId) {
-        var tempPositionArray = makePositionArray();
-        var origPageIndex = tempPositionArray[blockId].page;
+        var previousBlockPositions = makeBlockPositionMap();
+        var origPageIndex = previousBlockPositions[blockId].pageIndex;
         var newBlock = angular.copy(
           $scope.conference.registrationPages[origPageIndex].blocks[
-            tempPositionArray[blockId].block
+            previousBlockPositions[blockId].blockIndex
           ],
         );
-        var newPosition = tempPositionArray[blockId].block + 1;
+        var newPosition = previousBlockPositions[blockId].blockIndex + 1;
         newBlock.id = uuid();
         newBlock.profileType = null;
         newBlock.position = newPosition;
@@ -330,20 +329,20 @@ angular
           return;
         }
 
+        var previousBlockPositions = makeBlockPositionMap();
         if (growl) {
-          var t = makePositionArray();
           var block =
-            $scope.conference.registrationPages[t[blockId].page].blocks[
-              t[blockId].block
-            ];
+            $scope.conference.registrationPages[
+              previousBlockPositions[blockId].pageIndex
+            ].blocks[previousBlockPositions[blockId].blockIndex];
           var message = '"' + block.title + '" has been deleted.';
           GrowlService.growl($scope, 'conference', $scope.conference, message);
         }
 
-        var tempPositionArray = makePositionArray();
         _.remove(
-          $scope.conference.registrationPages[tempPositionArray[blockId].page]
-            .blocks,
+          $scope.conference.registrationPages[
+            previousBlockPositions[blockId].pageIndex
+          ].blocks,
           { id: blockId },
         );
       };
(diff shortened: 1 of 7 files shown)


wrandall22 (changes requested):
For the most part this looks really good and adds some good code coverage, thanks!

Some things to note:
1) You asked about commits, some things you could have done, which may have been slightly nicer, would have been to put the commit "Remove redundant GET request" after the test commits (TDD principle). You could also have combined the cleanup commits for `eventForm.js` into a single commit and have the description be a bulleted list of the cleanup actions done.

2) There are a few places where the test sentencing is a bit weird, I believe you did this to avoid putting a `describe` block around a single `it` block but retain the function name. There are probably trade-offs to readability and useless boilerplate here. I'm ok either way on it, but it does seem a tad jarring to read.

3) There are quite a few places (new and old) where tests are referencing hard-coded UUIDs and text. We should really work towards removing this habit from the codebase and use the reference locations of these values instead. Otherwise, if we change a value in the `testData.spec.js` file we'll have to search and replace it all over the codebase instead of it just working properly.

test/spec/controllers/eventForm.spec.js
```diff
@@ -0,0 +1,309 @@
...
+
+      expect(modalMessage.error).toHaveBeenCalledTimes(1);
+      const errorMessage = modalMessage.error.calls.argsFor(0)[0].message;
+
+      expect(errorMessage).toContain('<li>Question</li>');
+    });
+  });
+
+  it('addNewPage adds a new page', () => {
+    scope.addNewPage();
```
dr-bizz, line 290: Should you add a expect the pages to have a legnth of 2 first, so we know a page was added.

test/spec/controllers/eventRegistrations.spec.js
```diff
@@ -25,37 +23,236 @@ describe('Controller: eventRegistrations', function () {
...
+  });
+
+  it('toggleColumn toggles the column visibility', () => {
+    const setItem = spyOn($window.localStorage, 'setItem');
+    scope.toggleColumn(0);
+
+    expect(scope.blocks[0].visible).toBe(true);
+    expect(setItem).toHaveBeenCalledWith(
+      `visibleBlocks:${testData.conference.id}`,
+      '["e088fefc-eb9c-4904-b849-017facc9e063"]',
```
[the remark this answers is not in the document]
dr-bizz: Maybe make the uuid into a const which we export from testData.spec.js?

test/spec/controllers/eventRegistrations.spec.js
```diff
@@ -25,37 +23,236 @@ describe('Controller: eventRegistrations', function () {
...
+        Group: true,
+        GroupId: false,
+        Started: true,
+        Completed: true,
+      });
+    });
+  });
+
+  it('toggleColumn toggles the column visibility', () => {
+    const setItem = spyOn($window.localStorage, 'setItem');
```
dr-bizz, line 135: Would it be worth testing multiple columns being toggled?

test/spec/controllers/eventRegistrations.spec.js
```diff
@@ -25,37 +23,236 @@ describe('Controller: eventRegistrations', function () {
...
+    });
+
+    it('returns "" if the answer does not exist', () => {
+      scope.queryParameters.orderBy = 'last_name';
+
+      expect(scope.answerSort(testData.registration.registrants[0])).toBe('');
+    });
+
+    it('returns the value of simple answers', () => {
+      scope.queryParameters.orderBy = '9b83eebd-b064-4edf-92d0-7982a330272a';
```
dr-bizz, line 217: As Bill was mentioning, we should prob make this into a variable, so it's easier to read and not mess up the test.

test/spec/controllers/eventRegistrations.spec.js
```diff
@@ -25,37 +23,236 @@ describe('Controller: eventRegistrations', function () {
...
+
+      expect(scope.answerSort(testData.registration.registrants[0])).toBe(
+        'Test,Person',
+      );
+    });
+
+    it('returns the keys of checkbox answers', () => {
+      scope.queryParameters.orderBy = '18ccfb09-3006-4981-ab5e-405ccf2aad1c';
+
+      expect(scope.answerSort(testData.registration.registrants[0])).toBe(
```
dr-bizz, line 233: Should it return either it was true or false as well?

dr-bizz: I guess it doesn't need to as ir's just ordering the answers

test/spec/controllers/eventRegistrations.spec.js
```diff
@@ -150,4 +348,223 @@ describe('Controller: eventRegistrations', function () {
...
+
+      $httpBackend.flush();
+
+      expect(registrant.checkedInTimestamp).toBeNull();
+    });
+  });
+
+  describe('deleteRegistrant', () => {
+    it('deletes a registrant in a group', () => {
+      $httpBackend.expectDELETE(/^registrants\/.+$/).respond(204, '');
```
dr-bizz, line 536: Can we do a test where there are more than just one registrant in a group when one is deleted? This is linked to my comment about the error when deleting a registrant in the modal.

dr-bizz: Maybe this is doing that. But the error I had is around the same code

test/spec/controllers/reviewRegistration.spec.js
```diff
@@ -40,6 +186,50 @@ describe('Controller: ReviewRegistrationCtrl', function () {
...
+      it('is false on completed registrations', () => {
+        testData.conference.registrantTypes[1].acceptPayOnSite = true;
+        testData.registration.completed = true;
+        initController();
+
+        expect(scope.acceptedPaymentMethods().acceptPayOnSite).toBe(false);
+      });
+    });
+
+    it('returns false when no payment methods are accepted', () => {
```
dr-bizz, line 218: Is this the test when the form requires no payment?

dr-bizz (changes requested):

test/spec/controllers/reviewRegistration.spec.js
canac: Yes. Is there anything I can do to make that clearer?

app/scripts/controllers/eventRegistrations.js
```diff
@@ -573,43 +573,41 @@ angular
...
-            $http
-              .get('registrations/' + registrant.registrationId)
-              .then(function (response) {
-                var registration = response.data;
-                var url = 'registrations/' + registration.id;
-
-                if (registration.registrants.length > 1) {
-                  //Delete Registrant
-                  url = 'registrants/' + registrant.id;
+            const registration = $scope.getRegistration(
```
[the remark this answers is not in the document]
canac: Upon further testing, removing that GET request seemed to cause issues, so I reverted that change.

test/spec/controllers/eventRegistrations.spec.js
canac: Yes, this tests removing one registrant from a group with multiple registrants. I reverted the change to the controller, and that should fix the error.

app/scripts/controllers/eventForm.js
```diff
@@ -210,28 +208,29 @@ angular
           });
       };
 
-      var makePositionArray = function () {
-        var tempPositionArray = [];
+      // Generate a map of blockIds to their page and index within the page
+      var makeBlockPositionMap = function () {
+        var tempPositionArray = {};
```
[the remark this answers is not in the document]
canac: I fixed the variable name

test/spec/controllers/reviewRegistration.spec.js
dr-bizz: No you're good, I think I was just confirming

dr-bizz (approved):
Thank you for adding this coverage!

canac:
@wrandall22 Do you want to take another look, or should I dismiss your review?

PR #853, opened 2024-04-05
[HS-1132003] Upgrade rollbar

canac:
Upgrade rollbar to get additional telemetry such as network activity like `give-web` has. [Here's an example](https://app.rollbar.com/a/Cru/fix/item/ert-web/795/occurrence/402828749624#detail) of an occurrence in staging. Please double-check that all PII is removed.

Assisting with debugging https://secure.helpscout.net/conversation/2560557180/1132003?viewId=320673

### DIFF
--- a/app/scripts/errorNotify.js
+++ b/app/scripts/errorNotify.js
@@ -5,6 +5,7 @@ const rollbarConfig = {
   ignoredMessages: [
     "undefined is not an object (evaluating '__gCrWeb.autofill.extractForms')",
   ],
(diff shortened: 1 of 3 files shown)


wrandall22 (approved):

canac:
The credit card form input values were still being recorded as DOM interactions. All input values are now scrubbed. New occurrence with input scrubbing enabled: https://app.rollbar.com/a/Cru/fix/item/ert-web/795/occurrence/402830618045#detail. We could make the scrubbing more granular in the future if desired.

PR #854, opened 2024-05-01
[EVENT-845] Disable editing staff profile info

canac:
Prevent staff from editing name and email profile fields. These fields are prepopulated and should not be modified.

If you need a conference to test with, you're welcome to use this one: https://stage.eventregistrationtool.com/eventDetails/ec6ac5f9-1df0-4b69-b09d-f85ce39001ef. It is configured to allow Okta and Google logins.

https://jira.cru.org/browse/EVENT-845

### DIFF
--- a/app/scripts/directives/blocks.js
+++ b/app/scripts/directives/blocks.js
@@ -25,9 +25,10 @@ angular.module('confRegistrationWebApp').directive('nameQuestion', function () {
         $scope.answer.value = {};
       }
 
-      $scope.lockedStaffProfileBlock =
-        !!$rootScope.globalUser().employeeId &&
-        $scope.block.profileType === 'NAME';
+      const user = $rootScope.globalUser();
+      $scope.lockedStaffProfileBlock = Boolean(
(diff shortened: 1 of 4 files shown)


app/scripts/directives/blocks.js
```diff
@@ -20,10 +20,14 @@ angular.module('confRegistrationWebApp').directive('nameQuestion', function () {
...
     templateUrl: nameQuestionTemplate,
     restrict: 'E',
-    controller: function ($scope) {
+    controller: function ($rootScope, $scope) {
       if (!$scope.answer && !$scope.answer.value) {
         $scope.answer.value = {};
       }
+
+      $scope.lockedStaffProfileBlock =
+        !!$rootScope.globalUser().employeeId &&
```
wrandall22, line 29: Any chance of `rootScope.globalUser()` returning undefined and throwing an error here?

canac: Good catch. That is a possibility if this gets called before the profile finishes loading.

wrandall22 (approved):

PR #856, opened 2024-05-10
[MPDX-845] Allow editing non-primary registrant's profiles

canac:
A HelpScout user was unable to edit their spouse's name/email because #854 disables editing all registrants' profile fields ([ticket](https://secure.helpscout.net/conversation/2592103865/1152549?viewId=7296729)). This PR relaxes that restriction to only lock the primary registrant's profile.

I also trimmed down the test cases. We don't really need an exponential number of test cases for each condition, just one test case where all the `lockedStaffProfileBlock` conditions are true and one for each of the cases where a condition is false.

If you need a conference to test with, you're welcome to use [this one](https://stage.eventregistrationtool.com/register/ec6ac5f9-1df0-4b69-b09d-f85ce39001ef). It supports registrant groups.

https://jira.cru.org/browse/EVENT-845

### DIFF
--- a/app/scripts/directives/blocks.js
+++ b/app/scripts/directives/blocks.js
@@ -30,6 +30,8 @@ angular.module('confRegistrationWebApp').directive('nameQuestion', function () {
         !$scope.adminEditRegistrant &&
           user &&
           user.employeeId &&
+          $scope.currentRegistrant ===
+            $scope.currentRegistration.primaryRegistrantId &&
           $scope.block.profileType === 'NAME',
       );
     },
@@ -78,6 +80,8 @@ angular
           !$scope.adminEditRegistrant &&
             user &&
             user.employeeId &&
+            $scope.currentRegistrant ===
+              $scope.currentRegistration.primaryRegistrantId &&
             $scope.block.profileType === 'EMAIL',
         );
       },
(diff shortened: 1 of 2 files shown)


wrandall22:
This, then, will allow modifying name and email on secondary registrants that are staff. Is this problematic?

canac:
@wrandall22 Good question. The impetus for this change was to prevent spouses from logging in and changing the registrant's name to their spouse's name (https://secure.helpscout.net/conversation/2583427606/1145486/). Using Andy's example, if Catherine registers and is forced to keep her name as the primary registrant, it seems like she would put Andy's name as the spouse, which is what we want.

Also, the server auto-populates the name and email profile fields, and seems to only do that for the first registrant. We need to give the user some way to edit the blank profile fields for the secondary registrants.

wrandall22 (approved):

dr-bizz:
So, I tried to replicate this issue for a while but couldn't. I now know it's because I was logging in with Google instead of Okta.

Do you know If we also get the EmployeeId from Google or if that it only comes from OKTA?

canac:
@dr-bizz I thought that when I tried it before it would set my `employeeId` when I logged in with Google. But either I was mistaken or it's not happening anymore because I now have to login with Okta to get an employee id in my profile.

FYI, I have trouble getting the `employeeId` in my profile in staging even when logged in with Okta. Lee made a database change a few times that worked temporarily, but it would keep getting overwritten on the server side and come back as `null`.

PR #859, opened 2024-10-22
[HS-1245681] Hiding child question blocks

caleballdrin:
https://secure.helpscout.net/conversation/2738597988/1245681?folderId=7669074

Hide questions if it has a rule and the question that the rule is based on is hidden.


test/spec/directives/block.spec.js
```diff
@@ -47,6 +47,16 @@ describe('Directive: blocks', () => {
...
+      it('is false when currentRegistration is null', () => {
+        $scope.currentRegistration = null;
+        globalUserSpy.and.returnValue(null);
+
+        $compile('<name-question></name-question>')($scope);
+        $scope.$digest();
+
+        expect($scope.lockedStaffProfileBlock).toBe(false);
+      });
+
```
caleballdrin, line 59: @canac I'm not sure how these other 2 commits got in here. Do you think I just never pushed them from a long time ago?

test/spec/testData.spec.js
```diff
@@ -411,6 +411,157 @@ angular.module('confRegistrationWebApp').service('testData', function () {
           },
         ],
       },
+      {
```
caleballdrin, line 414: @canac Do you think I need to add `conference` to line 9 on [pageRegistration.html](https://github.com/CruGlobal/conf-registration-web/blob/44447f09d06ece3cceaccf3e00bac31c2e92adff/app/views/components/pageRegistration.html) ng-`show="blockVisible(block)"`

canac: I don't think so. I'm not fully sure how, but it appears that `pageRegistration.html` is connected to the `blockRegistration` controller, which sets up `$scope.blockVisible`. That's the function that the template is calling.

wrandall22:
For PRs that are from HS tickets, I prefer we use the `HS-123456` instead of `[No-Jira]` as a prefix in the title. This is more explanatory.

app/scripts/services/validateRegistrant.js
```diff
@@ -73,6 +82,19 @@ angular
...
+            if (
+              conference &&
+              !blockVisibleRuleCheck(
+                parentBlock,
+                registrant,
+                ruleType,
+                conference,
+              )
+            ) {
+              return false;
```
canac, line 96: Returning false makes `_.forEach` stop iterating.

```suggestion
              return;
```

canac (approved):
Let's ship it! 🚀

PR #860, opened 2024-10-22
[HS-1241612] Prevent conference names from containing special characters

canac:
# Description

* Prevent users from creating conferences with names containing ampersand or double quotes. Apparently, those special characters are causing issues on the server.
* Also prevent users from editing conferences to contain an ampersand or double quote. If they have an existing conference with an ampersand or a double quote, they will change the name before they can edit any other conference details. My understanding is that that is desirable.
* Also prevent users from editing conference abbreviations to contain an ampersand or double quote.

# Testing

* From the admin dashboard, click "Create New Event"
  * Validate that there are errors when the event name is empty (after focusing the name input and then moving focus) and the form cannot be submitted.
  * Validate that there are errors when the event name contains an ampersand (&) and the form cannot be submitted.
  * Validate that there are errors when the event name contains a double quote (") and the form cannot be submitted.
  * Validate that there are no errors when the event name is valid and the form can be submitted.
* From the admin dashboard, click on an event, then go to Details and change the Event Name.
  * Validate that there are errors when the event name contains an ampersand (&).
  * Validate that there are errors when the event name contains a double quote (").
  * Validate that there are no errors when the event name is valid and that the event name is updated.
* From the admin dashboard, click on an event, then go to Details and change the Event Abbreviation.
  * Validate that there are errors when the event abbreviation contains an ampersand (&).
  * Validate that there are errors when the event abbreviation contains a double quote (").
  * Validate that there are no errors when the event abbreviation is valid and that the event abbreviation is updated.

https://secure.helpscout.net/conversation/2731356723/1241612/

cc @hlbraddock

### DIFF
--- a/app/scripts/controllers/eventDashboard.js
+++ b/app/scripts/controllers/eventDashboard.js
@@ -41,10 +41,6 @@ angular
             templateUrl: createEventModalTemplate,
           })
           .result.then(function (conferenceName) {
-            if (!conferenceName) {
-              return;
-            }
-
             ConfCache.create(conferenceName).then(function (conference) {
               $location.path('/eventDetails/' + conference.id);
             });
--- a/test/spec/controllers/eventDetails.spec.js
+++ b/test/spec/controllers/eventDetails.spec.js
@@ -279,6 +279,22 @@ describe('Controller: eventDetails', function () {
         'Please enter which Event Type',
       );
     });
+
+    it('saveEvent() should validate the Event Name', () => {
+      scope.conference.name = 'Men & Women Conference';
+      scope.saveEvent();
+
+      expect(scope.notify.message.toString()).toContain(
+        'Please remove double quotes (") and ampersands (&) from the event name.',
+      );
+
+      scope.conference.name = '"Cru" Conference';
+      scope.saveEvent();
+
+      expect(scope.notify.message.toString()).toContain(
+        'Please remove double quotes (") and ampersands (&) from the event name.',
+      );
+    });
   });
 
   describe('Conference (Cru event) without ministry hosting', function () {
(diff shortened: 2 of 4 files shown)


app/scripts/controllers/eventDashboard.js
```diff
@@ -41,10 +41,6 @@ angular
             templateUrl: createEventModalTemplate,
           })
           .result.then(function (conferenceName) {
-            if (!conferenceName) {
-              return;
-            }
-
```
dr-bizz, line 47: Are you removing this since the conference name is always required?

test/spec/controllers/eventDetails.spec.js
```diff
@@ -268,16 +268,50 @@ describe('Controller: eventDetails', function () {
...
 
-      expect(scope.notify.message.toString()).not.toContain(
-        'Please enter which Event Type',
-      );
+        expect(scope.notify.message.toString()).not.toContain(
+          'Please enter which Event Type',
+        );
+      });
+
+      it('should validate the Event Name', () => {
```
dr-bizz, line 284: We might have these tests, but if not we should add them. Test to make sure the error messages do not show when there is no `&` or `"` in the name.

dr-bizz (approved):
It works well. Thank you or fixing this. I left a few comments, but I will approve so not to block you.

app/scripts/controllers/eventDashboard.js
canac: Yes

test/spec/controllers/eventDetails.spec.js
canac: Good idea, I'll add them.

PR #861, opened 2024-10-29
[EVENT-849] Always show promo code box

canac:
Users are requesting that the promo code box is always visible instead of requiring the user to click the "Add Promotion Code" link first.

HelpScout ticket: https://secure.helpscout.net/conversation/2718660687/1234291/

[EVENT-849](https://jira.cru.org/browse/EVENT-849)

### DIFF
--- a/app/views/reviewRegistration.html
+++ b/app/views/reviewRegistration.html
@@ -344,31 +344,30 @@
             </tr>
           </table>
 
-          <a
-            href=""
-            ng-click="showPromoInput = true"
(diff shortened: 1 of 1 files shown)


dr-bizz (commented):
The code looks great! Is there a way to test this? I try on my local but I don't have an event which has a payment details filled out and I don't know the test Merchant Account ID, to be able to test this.

dr-bizz (approved):
Works well!

PR #866, opened 2024-11-08
[no-Jira] Upgrade to webpack v5 and karma v6

canac:
Upgrade to webpack v5. Each step of the upgrade is in an atomic step that fixes a specific thing. After the webpack upgrade, an upgrade to the latest version of karma was required. I expect this PR to resolve a large number of the dependabot vulnerabilities, and then we can go from there to see the severity levels of what is left.

I manually merged this branch into staging.

Notable changes:

* Upgrade to node v18 to get webpack running
* Migrate deprecated webpack plugins to newer versions
* Migrate from deprecated `@import` to `@use` in SCSS
* Apparently due to changes in bundling, `testData` cannot be accessed outside of an `describe`, `it`, or `before*` block, so some shared variable initializations are moved inside of `beforeEach` blocks
* Fixed tests from #859 that were silently not running because of an exception in the variable initialization due to the answer block ids being incorrect
* Fix errors caused by `describe` blocks without any tests inside of them by adding a trivial test. Those tests could be expanded in the future but feels out of scope for this PR, and I didn't want to just delete those test files.


package.json
```diff
@@ -94,7 +94,6 @@
     "webpack-bundle-analyzer": "^4.10.2",
     "webpack-cli": "^5.1.4",
     "webpack-dev-server": "^5.1.0",
-    "webpack-inline-manifest-plugin": "^4.0.1",
```
wrandall22, line 97: It would be good to have in the commit message why this is no longer needed.

webpack.config.js
```diff
@@ -164,25 +165,18 @@ module.exports = (env = {}) => {
...
+          generator: {
+            filename: '[name].[contenthash][ext]',
+          },
         },
-
         {
           test: /\.(svg|png|jpe?g|gif)/,
+          type: 'asset/resource',
+          generator: {
+            filename: '[name].[contenthash][ext]',
```
dr-bizz, line 177: I assume you no longer need to add  the dot between the contenthash and ext

webpack.config.js
```diff
@@ -2,13 +2,13 @@ const webpack = require('webpack');
...
-  require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
-const HtmlWebpackExcludeAssetsPlugin = require('html-webpack-exclude-assets-plugin');
-const WebpackInlineManifestPlugin = require('webpack-inline-manifest-plugin');
-const WebappWebpackPlugin = require('webapp-webpack-plugin');
-const SriPlugin = require('webpack-subresource-integrity');
+const {
+  HtmlWebpackSkipAssetsPlugin,
+} = require('html-webpack-skip-assets-plugin');
```
dr-bizz, line 9: Good find!

app/scripts/services/httpUrlInterceptor.js
```diff
@@ -3,7 +3,11 @@ angular
   .factory('httpUrlInterceptor', function (envService) {
     return {
       request: function (config) {
-        var passthroughRegexs = [/https?:\/\/.*/, /^views\/.*/, /template\/.*/];
+        var passthroughRegexs = [
+          /https?:\/\/.*/,
+          /^\/views\/.*/,
```
dr-bizz, line 8: What was the cause for this update?

dr-bizz (commented):
I haven't found much yet, but here are a few comments

app/scripts/services/httpUrlInterceptor.js
canac: The URL used to load the view HTML files changed from `views/...` to `/views/...`. Probably because of an update to `ngtemplate-loader`.

app/styles/common.scss
```diff
@@ -1,11 +1,15 @@
+@use '~bootstrap-sass/assets/stylesheets/bootstrap';
+@use 'variables';
+@use 'typography';
```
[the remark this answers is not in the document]
canac: > Are you using `@use 'typography';` in this file? I might have missed it if you are.

`@extend .sans-serif;` and similar is a reference to `typography.scss`.

> I guess the typography file is just added classes, but then why was it removed from style.scss?

Before `style.scss` `@import`ed everything once into the global scope. `@import` is deprecated now and we have to `@use`. `@use` only makes the variables and styles from that module available in the current file. So now we have to to `@use` `variables` and `typography` in every file that references them. There's also no reason to have the import in `style.scss` anymore because it doesn't actually use any of the variables and styles that `variables.scss` or `typography.scss` define. Hopefully that makes sense.

webpack.config.js
canac: Correct, it's not needed anymore

https://webpack.js.org/guides/asset-modules/#custom-output-filename

package.json
canac: Sounds good. I can't change the commit message without rebasing and force-pushing, which will make it harder to review. But I will make sure to do that before merging.

```
Remove webpack-inline-manifest-plugin

webpack-inline-manifest-plugin only supports webpack v4
and manifest inlining is not necessary with webpack v5 and
HtmlWebpackPlugin v5. Scripts and styles are automatically
included in the HTML file.
```

.tool-versions
```diff
@@ -1,2 +1 @@
-nodejs 16.10.0
```
dr-bizz, line 1: Why are we changing the `.tool-versions` so it doesn't handle the node version? It's default for asdf to use it for setting the node version

dr-bizz (approved):
This is looking great! I'm going to approve. I had one comment, but I don't want to block you

PR #867, opened 2024-11-08
HS-1245681 Fix Hidden Question Validation

caleballdrin:
https://secure.helpscout.net/conversation/2738597988/1245681?folderId=7669074

-In the `validate()` function we need to call `blockVisibleRuleCheck()` with `conference` so it can accurately check if the question is hidden.
-Add tests

### DIFF
--- a/app/scripts/services/validateRegistrant.js
+++ b/app/scripts/services/validateRegistrant.js
@@ -287,6 +287,7 @@ angular
               block,
               registrant,
               ruleTypeConstants.SHOW_QUESTION,
+              conference,
             ) ||
             !blockInRegistrantType(block, registrant)
           ) {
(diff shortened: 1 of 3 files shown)


app/scripts/services/validateRegistrant.js
```diff
@@ -287,6 +287,7 @@ angular
               block,
               registrant,
               ruleTypeConstants.SHOW_QUESTION,
+              conference,
```
canac, line 290: Should we add this to `this.checkboxDisable` also?

canac: That will be done in a follow-up PR

canac (approved):
Looks good! Thanks for working on this!

PR #871, opened 2024-11-12
[no-Jira] Dependabot semver minor upgrades

canac:
Upgrade all vulnerable packages that can be updated with a semver minor upgrade. This should resolve many of the remaining vulnerabilities but a handful are in transitive dependencies of packages without updates that fix the vulnerabilities.

I manually merged this branch into staging.


dr-bizz (approved):
Looks great!

PR #875, opened 2025-03-07
[no-Jira] Replace the update-staging workflow with merge-bot

canac:
The `update-staging` workflow successfully merges branches into staging, but it doesn't trigger a deployment because workflows can't trigger other workflows. This PR removes that workflow and enables the merge bot for this repo.


canac:
The workflow didn't trigger on this PR, but I'm thinking that maybe the config file has to be in `master` to work.

canac:
> The workflow didn't trigger on this PR, but I'm thinking that maybe the config file has to be in `master` to work.

That was it. The `On Staging` label works now for other PRs. 🎉

PR #876, opened 2025-03-07
[no-Jira] Add accessibility labels to registrant-visible fields

canac:
In this PR, I added accessibility labels to all form inputs that are visible to a user registering for a conference. Because `input`s are connected with their `label` correctly now, clicking on a label focuses the input. That is native browser behavior.

### DIFF
--- a/app/views/paymentMethods/creditCard.html
+++ b/app/views/paymentMethods/creditCard.html
@@ -1,71 +1,81 @@
 <div class="form-horizontal">
   <div class="form-group">
     <div class="col-sm-12">
-      <label translate>Name on Card</label>
-      <input
-        type="text"
-        ng-model="currentPayment.creditCard.nameOnCard"
-        class="form-control"
-      />
+      <label>
+        <span translate>Name on Card</span>
+        <input
+          type="text"
+          ng-model="currentPayment.creditCard.nameOnCard"
+          class="form-control"
+        />
+      </label>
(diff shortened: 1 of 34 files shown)


canac:
> The only thing I noticed, especially on group inputs, is that it announces the aria-label after the options (e.g. for campus opportunities it will announce the option then tell you what the option is for rather than announcing the question)

@rguinee Thanks for taking a look at this! I was thinking it was a good thing for the label to be the option and the group to be the question that the radio button answers. Otherwise if the user uses the arrow keys to move through the radio buttons it will announce the answer first every time, which seems redundant. I'm new to all this, so I appreciate hearing other opinions.

app/views/paymentMethods/creditCard.html
```diff
@@ -1,124 +1,147 @@
...
-      <label translate>Card Security Code</label>
-      <input
-        class="form-control"
-        type="text"
-        maxlength="4"
-        ng-model="currentPayment.creditCard.cvvNumber"
-        autocomplete="off"
-      />
+  <fieldset>
+    <legend class="sr-only">Credit Card</legend>
```
dr-bizz, line 3: Great idea

dr-bizz (approved):
Awesome work. I haven't looked at the UI, but from the code, this looks great!

rguinee (approved):

PR #877, opened 2025-03-17
[HS-1325929] Fix force selection checkbox being hidden

canac:
Fix DOM nesting error introduced in #876 that prevented the "Force selection" checkbox from displaying.

HelpScout ticket: https://secure.helpscout.net/conversation/2879252870/1325929?viewId=320670

## Testing

* Add a Checkbox question to the event
* Add an answer option
* Check the answer option you added
* Ensure that the Force selection checkbox is visible

### DIFF
--- a/app/views/blocks/checkboxQuestion.html
+++ b/app/views/blocks/checkboxQuestion.html
@@ -35,25 +35,25 @@
         >
       </label>
     </div>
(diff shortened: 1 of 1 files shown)


dr-bizz (approved):
Looks and tests great

PR #878, opened 2025-03-27
Upgrade to Node 22

dr-bizz:
## Description
In this PR, I have upgraded Node from version 18 to version 22. I have also upgraded Yarn to version 4. Yarn 4 has stricter rules around `--immutable-cache`, causing us to remove it from the GitHub Actions.

I have also updated the TS ESLint packages, as there was a warning message when running lint.

## Changes:
- `node`: `18` -> `22`
- `yarn`: `1` -> `4`
- `@typescript-eslint/eslint-plugin`: `4` -> `5`
- `@typescript-eslint/parser`: `4` -> `5`
- Removed `--immutable-cache` from GHA. _The way we could get this to pass is to track all .`yarn/cache` folder, but this would bloat the repo and the only need for the cache files would be for the GHA, as we use node_modules._


package.json
```diff
@@ -49,8 +49,8 @@
     "@types/react": "^17.0.14",
     "@types/react-bootstrap": "^0.32.31",
     "@types/react-dom": "^17.0.9",
-    "@typescript-eslint/eslint-plugin": "^4.33.0",
-    "@typescript-eslint/parser": "^4.33.0",
+    "@typescript-eslint/eslint-plugin": "5.57.0",
+    "@typescript-eslint/parser": "5.57.0",
```
canac, line 53:
```suggestion
    "@typescript-eslint/eslint-plugin": "^5.57.0",
    "@typescript-eslint/parser": "^5.57.0",
```

dr-bizz:
I will clean up the commit after approval is completed.

README.md
```diff
@@ -5,23 +5,50 @@ https://www.eventregistrationtool.com | https://stage.eventregistrationtool.com
...
-- `yarn run test` or `npm run test` to run karma tests
-- `yarn run lint` or `npm run lint` to run eslint
-- `yarn run build` or `npm run build` to generate minified output files. These files are output to `/dist`.
-- `yarn run build:analyze` or `npm run build:analyze` to open a visualization of bundle sizes after building
-- `yarn run angular-gettext-extract` or `npm run angular-gettext-extract` to extract strings to `languages/ert.pot` for translation
+- `yarn run test` to run karma tests
+- `yarn run lint` to run eslint
+- `yarn run build` to generate minified output files. These files are output to `/dist`.
+- `yarn run build:analyze` to open a visualization of bundle sizes after building
+- `yarn run angular-gettext-extract` to extract strings to `languages/ert.pot` for translation
```
canac, line 51: I prefer the shorter versions of these commands

```suggestion
- `yarn test` to run karma tests
- `yarn lint` to run eslint
- `yarn build` to generate minified output files. These files are output to `/dist`.
- `yarn build:analyze` to open a visualization of bundle sizes after building
- `yarn angular-gettext-extract` to extract strings to `languages/ert.pot` for translation
```

.github/workflows/ci.yml
```diff
@@ -60,18 +60,9 @@ jobs:
         with:
           node-version-file: .node-version
       - name: 📦 Install Dependencies
-        run: yarn install --immutable --immutable-cache
+        run: yarn install --immutable
       - name: 💅 Prettier Check
         run: yarn prettier:check
-  yarn-check-cache:
```
canac, line 66: The docs make it seem like `--check-cache` can still be useful to make sure that the packages in the registry match the checksums in the lockfile.

canac (commented):
Can you also update `.node-version`? That's the version that Netlify uses to build the site. Unfortunately, it's not smart enough to use `.tool-versions`, so we have to set it in both places.

Can we also update actions/checkout to v4? If I recall correctly, v2 was causing warnings about running on an old node version.

.github/workflows/ci.yml
dr-bizz: From what I read, it only checks packages that are stored in the cache folder. Since we don't have any packages in the cache folder, it doesn't do anything.

canac: On second thought, I think you're right. If an attacker commits a `yarn.lock` with an incorrect checksum, it will fail to install, even without `--check-cache`.

canac (approved):
Thanks for your work on this!

PR #879, opened 2025-08-13
Update asdf Node version to v24.3.0

zweatshirt:
- This PR attempts to update the project's asdf nodejs version to 24.3.0
- I did not notice any issues when running `yarn install`, but it is worth checking again by another set of eyes.


.tool-versions
```diff
@@ -1 +1 @@
-nodejs 22.14.0
```
dr-bizz, line 1: Let's use 24.3.0

dr-bizz (changes requested):
Good work! Please change the version to 24.3.0. That is the version we are using on a number of different projects. This makes it easier to manage when switching between projects.

Please also removed the yarn.lock file and run `yarn` to ensure all packages are install with `nodejs` version 24.

zweatshirt:
@dr-bizz Let me know what you think. I'd also like to try a new PR that replaces instanbul-instrumenter-loader with babel-plugin-instanbul and nyc (or a similar solution), since currently features like `??` are not supported.

dr-bizz (commented):
Please could you combine your commits so there are 2 commits. The first three commits can be combined, and the last two can be combined.

The other thing I noticed is that on my local, I've having issues installing the pckages on yarn. I am using yarn 4.7.0

zweatshirt:
@dr-bizz From looking at package.json the project also uses Yarn 4.7.0. I have not run into any errors when I run `yarn` or `yarn install` on 4.7.0 on my end but I do get warnings. I can spend some time trying to resolve that if needed.

PR #880, opened 2025-08-15
EVENT-855 Frontend - ERT Couple Type

zweatshirt:
[Jira ticket](https://jira.cru.org/browse/EVENT-855)

What is done:
- Deleting spouse-couple groups
- Withdrawing spouse-couple groups
- Notifying user via modal that a spouse-couple group will be deleted or withdrawn
- - Stopping admin from adding more than 1 spouse and the ability to add all other types. This is default ERT behavior and shouldn't be (but if it is still wanted it can be reverted).
- Stop people from changing type in the Add Registration Modal when adding a registrant from a group -- necessary to avoid changing spouse type but can be targeted to spouse only later if needed.
- Payment handling logic (contained within the Couple type -- the spouse should not be used at all for payments or refunds)

Other bugs resolved:
- The ability to register non-dependents in groups has been removed.
- In Registrant Types, the ability to make the primary a dependent has been removed. There are a few reasons I can see this being intended so I can revert this if needed. 

To finish:
- Tests (almost done locally)
- Potential UAT concerns

### DIFF
--- a/app/views/reviewRegistration.html
+++ b/app/views/reviewRegistration.html
@@ -172,7 +172,7 @@
       >
         Add a Group Member
       </h2>
-      <registration-type-select></registration-type-select>
+      <registration-type-select />
     </section>
     <section>
       <h2 class="page-title border" translate>Summary</h2>


dr-bizz (changes requested):
Really impressive work on this! It demonstrates that you have thought this through a lot.

I've made a few comments while reviewing the code and UI.

I also found 2 errors in the UI during testing.

**Error 1**: When deleting a `spouse` which was added to a `Default` registrant type, it comes up with the same message as a `spouse` linked to a `couple`.

Image

**Error 2**: I came across this error a few times, nearly always when registering. I was testing on my local, so it could be something todo with that.

Image

app/scripts/utils/coupleTypeUtils.js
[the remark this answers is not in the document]
zweatshirt: @dr-bizz I would actually prefer to use native functions and wasn't the biggest fan of using lodash, but was trying to keep code consistent. I'm thinking we should make a standard habit of using one over the other in the future. Let me know what you think.

app/scripts/directives/showGroupModal.js
```diff
@@ -23,37 +24,86 @@ angular.module('confRegistrationWebApp').component('showGroupModal', {
...
+      );
+      if (
+        !primaryRegistrantType ||
+        !primaryRegistrantType.allowedRegistrantTypeSet
+      ) {
+        return false;
       }
 
-      const registration = this.getRegistration(this.registrationId);
+      const allowedDependentsAmount = _.find(
```
dr-bizz, line 87: [Must-fix] Potential null pointer when _.find() returns undefined. Evidence: `_.find(...).numberOfChildRegistrants` without null check. Impact: Could crash UI when allowedRegistrantTypeSet doesn't contain matching childRegistrantTypeId. Fix: Add null check before accessing numberOfChildRegistrants. [Category: File]

app/scripts/controllers/eventRegistrations.js
```diff
@@ -563,59 +632,82 @@ angular
...
+
+            const reg = $scope.getRegistration(registrant.registrationId);
+            let url;
+            if (reg.registrants.length === 1 || isCouple) {
+              url = 'registrations/' + registration.id;
+            } else {
+              url = 'registrants/' + registrant.id;
+            }
+            $http.delete(url).then(function () {
+              $scope.refreshRegistrations();
```
dr-bizz, line 708: [Must-fix] Missing error handling in HTTP delete operation. Evidence: `$http.delete(url).then(...)` without .catch(). Impact: Failed deletions could leave UI in inconsistent state. Fix: Add .catch() handler. [Category: File]

app/scripts/controllers/registrationModal.js
```diff
@@ -131,5 +133,43 @@ angular
...
+            ) {
+              return false;
+            }
+
+            const hasAssociation = coupleType.allowedRegistrantTypeSet.some(
+              (association) => {
+                const isMatch = association.childRegistrantTypeId === type.id;
+
+                const hasChildRegistrants =
+                  association.numberOfChildRegistrants > 0;
```
dr-bizz, line 162: [Must-fix] Logic assumes numberOfChildRegistrants exists without validation. Evidence: `association.numberOfChildRegistrants > 0`. Impact: Runtime error if property is undefined. Fix: Add validation `association.numberOfChildRegistrants && association.numberOfChildRegistrants > 0`. [Category: File]

app/scripts/controllers/eventRegistrations.js
```diff
@@ -509,32 +524,86 @@
...
+          noString: noMessage || 'Cancel',
+          normalSize: true,
+        });
+      };
+
+      $scope.buildCoupleWithdrawMessage = function (value) {
+        const title = value
+          ? 'Withdraw Couple/Spouse'
+          : 'Reinstate Couple/Spouse';
+        const yesString = value ? 'Withdraw Both' : 'Reinstate Both';
```
dr-bizz, line 541: [Nice-to-have] Could extract modal confirmation pattern to reusable utility. Evidence: `showModal` function creates similar modals. Impact: Consistency and maintainability. Suggestion: Use existing modalMessage.confirm pattern. [Category: Pattern]

app/scripts/controllers/eventRegistrations.js
dr-bizz, line 708: [Must-fix] Missing error handling in HTTP delete operation. Evidence: `$http.delete(url).then(...)` without .catch(). Impact: Failed deletions could leave UI in inconsistent state. Fix: Add .catch() handler. [Category: File]

app/scripts/controllers/registrationModal.js
dr-bizz, line 162: [Must-fix] Logic assumes numberOfChildRegistrants exists without validation. Evidence: `association.numberOfChildRegistrants > 0`. Impact: Runtime error if property is undefined. Fix: Add validation `association.numberOfChildRegistrants && association.numberOfChildRegistrants > 0`. [Category: File]

dr-bizz (changes requested):
Ai Review

app/scripts/directives/showGroupModal.js
```diff
@@ -23,37 +24,86 @@ angular.module('confRegistrationWebApp').component('showGroupModal', {
...
+        primaryRegistrantType.allowedRegistrantTypeSet,
+        (set) => set.childRegistrantTypeId === type.id,
+      ).numberOfChildRegistrants;
```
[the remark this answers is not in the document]
zweatshirt: This property is used to determine the maximum number of dependents that can be registered in a primary's group. I also found the property name to be confusing so I'll change this code to make it clearer.

dr-bizz (commented):
Awesome work! Keep it up. This is super close to being ready for production

app/views/eventDetails/regTypes.html
```diff
@@ -108,117 +131,132 @@ <h4 translate>Registration Form</h4>
...
+                    ng-repeat="childType in type.allowedRegistrantTypeSet track by $index"
+                    ng-if="shouldShowChildType(childType, type)"
+                  >
+                    <div class="row">
+                      <div class="col-md-6">
+                        <label class="control-label text-left">
+                          <input
+                            type="checkbox"
+                            ng-model="type.allowedRegistrantTypeSet[$index].selected"
+                            ng-disabled="disableField('groupSubRegistrantType', type.defaultTypeKey, type.name)"
```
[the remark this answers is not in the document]
zweatshirt: Good catch. Let me look into this more and I'll get back to you about it.

app/scripts/controllers/eventRegistrations.js
```diff
@@ -563,59 +632,82 @@ angular
...
+        let title = 'Delete Registrant';
+        let yesString = 'Delete';
+        let warningMessage = 'Are you sure you want to delete this registrant?';
+        if (isCouple) {
+          title = 'Delete Couple/Spouse Registrants';
+          yesString = 'Delete Both';
+          warningMessage =
+            'Deleting this registrant will also delete their spouse and any associated registrations.<br><br>' +
+            warningMessage;
+        }
```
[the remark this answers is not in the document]
zweatshirt: This is fixed, I'll be pushing it soon

test/spec/controllers/eventRegistrations.spec.js
```diff
@@ -594,11 +609,66 @@ describe('Controller: eventRegistrations', function () {
...
+        .expectGET(/^registrations\/.+$/)
+        .respond(200, testData.coupleRegistration);
+
+      $httpBackend
+        .expectDELETE('registrations/709738ff-da79-4eed-aacd-d9f005fc7f4e')
+        .respond(204, '');
+
+      const registrant = scope.registrants[0];
+      scope.deleteRegistrant(registrant);
+
```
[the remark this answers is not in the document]
zweatshirt: The behavior is that one registrant is passed to delete, we check if it is a couple or spouse, and we delete the whole registration if it is a couple-spouse pair (deleting both registrants). This assumes that the registration will always only contain the couple and spouse, which I believe is correct because they should always be in a pair. 

I'm adding `expect(scope.registrants.length).toBe(2)` before the delete to make this more apparent

app/scripts/controllers/eventRegistrations.js
zweatshirt: Currently not a critically needed change, I will pass this for now unless you prefer I do it

app/views/eventDetails/regTypes.html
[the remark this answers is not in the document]
zweatshirt: I've tried to resolve this. It was there initially when the original decision was to no longer use the default spouse type.

zweatshirt:
@dr-bizz Apologies for the delay with this. There is a chance I missed a couple comments so please let me know. Some of the AI ones in particular I'm not sure I always agree with. There was a lot that was done and some stuff that seems out of scope such as fixing `allowGroupRegistration` which was being evaluated in a directive by its reference rather than its return value. Because this was a production bug, it's challenging to tell what the intended behavior is.

test/spec/controllers/showGroupModal.spec.js
```diff
@@ -0,0 +1,121 @@
...
+
+      expect(controller.registrationTypeFull({ id: 'any-type-id' })).toBe(
+        false,
+      );
+    });
+
+    it('should return false when allowedRegistrantTypeSet is undefined', function () {
+      controller.getRegistrantType.and.returnValue({
+        id: 'primary-type-id',
+        allowedRegistrantTypeSet: null,
```
dr-bizz, line 63:
```suggestion
        allowedRegistrantTypeSet: undefined,
```

dr-bizz, line 63:
```suggestion
        allowedRegistrantTypeSet: undefined,
```

dr-bizz (approved):
🔥🔥 Amazing work on this PR 💯💯

I have left a few comments, but none of them will stop the rollout, just very minor things.

If you can fix those comments and then get your comments down from 60 to around 15-20, this will be ready for production. 

Again, fantastic work on this and persevering through all my comments and changes from stakeholders.

PR #881, opened 2025-08-19
Updated cru-payments to the latest version

dr-bizz:
## Description
In this PR, I update the cru-payments package to the latest version that is built with NodeJS 24 and fixes some security issues.

## Testing
- Log into ERT Staging
- Sign up for an event
- Enter incorrect CC details to ensure alerts are triggered.
- Enter correct CC details to ensure payment is successful.


canac (approved):
Looks great!

PR #883, opened 2025-08-20
Revert react-bootstrap due to styling issues

zweatshirt:
- **Please refer to the `promoUpload` and `journalUpload` pages as a reference to why this PR is necessary as well as PR #879.**
- Updating react-bootstrap affected styling in ways that was initially hard to catch.
- This attempts to revert react-bootstrap back to the original, while trying to satisfy `lint:ts`.
- `react-bootstrap.d.ts` is only used here to silence the linter in regards to imported react-bootstrap components and should be removed if there are any updates to the project react-bootstrap dependency in the future.


dr-bizz:
I just realised we don't have a test for build

PR #884, opened 2025-08-21
CI Build workflow: Adds build check to CI workflow

zweatshirt:


.github/workflows/ci.yml
```diff
@@ -63,6 +63,18 @@ jobs:
         run: yarn install --immutable
       - name: 💅 Prettier Check
         run: yarn prettier:check
+  build:
+    runs-on: ubuntu-latest
```
dr-bizz, line 67: Your code is great. Can we also add icons to their names?
```
name: 🏗️ Build app
```

This is one of the CI files I like to reference: https://github.com/CruGlobal/give-web/blob/bf5a90d45a79961e57e788ad869afe765ed568d5/.github/workflows/build.yml

_(I think I sent you to MPDX, when I meant to say Give)_

dr-bizz (changes requested):
Everything works great. If you could add icons to the name, it would give it a nice look.

zweatshirt:
@dr-bizz Due to the name additions, branch protection rules will need to be updated. Probably best to do that just before merging this.

PR #885, opened 2025-08-29
[no-Jira] Add a link to authenticate localhost from staging

canac:
Streamline local development by adding a link to staging that passing the staging auth token to the local dev server.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,8 @@ permissions:
 
 jobs:
   test:
+    name: ✅ Tests
     runs-on: ubuntu-latest
-    name: tests
     steps:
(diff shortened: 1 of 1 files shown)


canac:
> Will there be backend work needed to allow localhost to connect and authenticate without the need for staging?

@dr-bizz Yes, that is my understanding. When authenticating from localhost, the API sets the Okta `redirect_uri` to https://api.stage.eventregistrationtool.com/eventhub-api/rest/auth/okta/login and then redirects to the staging ERT instead of localhost. If it were in Rails, maybe I'd try to dive into the backend, but I don't really want to touch Java 😄. I went with this solution because it's the simplest one I could think of without needing to change the API code.

PR #889, opened 2025-09-17
Event-869 - Fix radio button persistence and prevent tracking hidden registrant type questions

dr-bizz:
## Description
In this PR, I've fixed the radio button selected value when editing choices, so it persists, and ensured that we don't save questions/answers that the admin has asked us not to track.

 ## Changes Made
  - **Radio button value persistence**: When editing a radio choice value that is
  currently selected, the selection now properly updates to the new value instead of
  becoming unselected
  - **Registrant type visibility**: Refactored `blockInRegistrantType` to
  `shouldShowForRegistrantType` for better clarity and made it public for testing and use across ERT.
  - **Answer initialization**: Added logic to only initialize answers for blocks that
  should be visible to the specific registrant type
  - **Test coverage**: Added comprehensive tests for the `shouldShowForRegistrantType`
  function

  ## Test Plan
- [ ] Radio button maintains selection when choice value is edited (option2 question)
- [ ] Ensure the question has a default value.
- [ ] Test the registration to ensure only relevant registrant types see appropriate questions. All other questions are shown as usual.
- [ ] When you export the registrations, the questions that didn't show up on the registration don't have the default value if they were not filled out.

### DIFF
--- a/app/scripts/directives/blockRegistration.js
+++ b/app/scripts/directives/blockRegistration.js
@@ -73,7 +73,7 @@ angular
           // Since we render all blocks for all registrants, even if they are not for this registrant type.
           // We ONLY need to initialize the answers/questions for the blocks that are visible
           const weShouldInitializeAnswer =
-            validateRegistrant.blockInRegistrantType(
+            validateRegistrant.shouldShowForRegistrantType(
               $scope.block,
               $scope.currentRegistration.registrants[registrantIndex],
             );
--- a/app/scripts/services/validateRegistrant.js
+++ b/app/scripts/services/validateRegistrant.js
@@ -210,9 +210,9 @@ angular
         );
       };
 
-      // We store the registrantTypes on the block that should NOT show the question.
-      // Which is why the value is inverted here.
-      this.blockInRegistrantType = (block, registrant) => {
+      this.shouldShowForRegistrantType = (block, registrant) => {
+        // We store the registrantTypes on the block that should NOT show the question.
+        // Which is why the value is inverted here.
         return !_.includes(block.registrantTypes, registrant.registrantTypeId);
       };
 
@@ -247,7 +247,7 @@ angular
             ruleTypeConstants.SHOW_QUESTION,
             conference,
           ) &&
-          this.blockInRegistrantType(block, registrant) &&
+          this.shouldShowForRegistrantType(block, registrant) &&
           this.isAnyChoiceVisible(block, registrant);
 
         return block.adminOnly && !isAdmin ? false : visible;
@@ -291,7 +291,7 @@ angular
               ruleTypeConstants.SHOW_QUESTION,
               conference,
             ) ||
-            !this.blockInRegistrantType(block, registrant)
+            !this.shouldShowForRegistrantType(block, registrant)
           ) {
             return;
           }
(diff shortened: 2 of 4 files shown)


app/scripts/directives/blockRegistration.js
```diff
@@ -70,6 +70,18 @@ angular
             return;
           }
 
+          // Since we render all blocks for all registrants, even if they are not for this registrant type.
+          // We ONLY need to initialize the answers/questions for the blocks that are visible
+          const weShouldInitializeAnswer =
+            validateRegistrant.blockInRegistrantType(
```
canac, line 76: This method doesn't seem to exist, and appears to be breaking staging.

app/scripts/directives/blockEditor.js
```diff
@@ -382,6 +392,33 @@ angular.module('confRegistrationWebApp').directive('blockEditor', function () {
...
+          // If the previous value matches the currently selected value,
+          // update the selected value to the new edited value
+          if (previousValue && previousValue === currentSelectedValue) {
+            const editedOption = $scope.block.content.choices[optionIndex];
+            $scope.block.content.default = editedOption.value;
+
+            // Also update the answer model
+            if ($scope.answer) {
+              $scope.answer.value = editedOption.value;
+              $scope.answer.hashKey = editedOption;
```
canac, line 415: Is setting the hasKey is necessary?

app/scripts/directives/blocks.js
```diff
@@ -169,6 +169,28 @@ angular
           ['selectedAnswer', 'otherAnswer'],
           updateAnswerValueDebounced,
         );
+
+        // Watch for changes to answer.value and update selectedAnswer accordingly
+        $scope.$watch('answer.value', function (newValue, oldValue) {
```
canac, line 174: This logic looks similar to this section. Could they be combined?

```js
const answerValue = $scope.answer ? $scope.answer.value : '';
if (
  $scope.block.content.otherOption &&
  $scope.block.content.otherOption.enabled &&
  answerValue &&
  !_.includes(_.map($scope.block.content.choices, 'value'), answerValue)
) {
  // An answer is present that isn't in the list of available choices, so it is an "Other" answer
  $scope.selectedAnswer = $scope.otherSentinel;
  $scope.otherAnswer = answerValue;
} else {
  $scope.selectedAnswer = answerValue;
  $scope.otherAnswer = '';
}
```

app/scripts/services/validateRegistrant.js
```diff
@@ -210,7 +210,9 @@ angular
         );
       };
 
-      const blockInRegistrantType = (block, registrant) => {
+      this.shouldShowForRegistrantType = (block, registrant) => {
+        // We store the registrantTypes on the block that should NOT show the question.
```
canac, line 214: This reads clearer to me. I also did not know this until today. Very unintuitive!

```suggestion
        // block.registrantTypes contains the ids of registrants that should NOT show the question.
```

canac (approved):
This is tricky code, but it seems to fix the bug you described. Thanks for noticing it and fixing it!

app/scripts/directives/blockEditor.js
dr-bizz: Good catch, we don't need it

app/scripts/directives/blocks.js
dr-bizz: Great simplification! I'll make the change

PR #890, opened 2025-09-19
Resolve hydration issue for eventOverview

zweatshirt:
- Updates conference registration count in eventOverview for hydration. Users would have to reload the page previously.


dr-bizz (approved):
Thank you for adding this and cleaning up the codebase.

It might be worth adding a test, but don't spend too much time on it.

PR #892, opened 2025-09-26
[no-Jira] Remove .node-version to use .tool-versions

canac:


dr-bizz (approved):
Is there a way we can use one file to set the node version?

For future devs, this might not be obvious.

Approving, so I'm not blocking

PR #982, opened 2026-04-24
[No Jira] KR5: Receive school location data from Connections

kegrimes:
## Description

We want to fetch school data from SalesForce, the source of truth, to prevent duplicate student records due to mismatching schools. 

- Connect `/campuses/connections/search` - returns `{ id, name }`
- Ensure all future campus questions use new API while supporting old API as well

More context is found in these Slack threads:
https://cru-main.slack.com/archives/GG4669AH5/p1768512017224959 
https://cru-main.slack.com/archives/GG4669AH5/p1770151342647119 

## Testing

Initial campus test:
- Go to event dashboard
- Create an event with the a campus question
- Go to the registration page for this event & open up DevTools --> Network
- Search for a campus
- Verify the list of campuses is appearing from the correct API (`/campuses/connections/search`)
- Click one and go to review registration page
- Click on your name and ensure the correct campus name appears
- Confirm the registration
- Refresh page and click on Profile cache --> verify you see the `connectionsSchoolId` populated with the correct school id

Already set campus test:
- Create an event with the a campus question
- Go to the registration page and ensure your campus is pre-populated

## Checklist


kegrimes (commented):
CAMPUS_V2 introduces a dedicated profile type that stores the Salesforce campus id while keeping V1 records intact. The HTTP interceptor centralizes the {id,name}→id normalization across all PUT/POST paths, and the display layer handles both shapes via fallback. Test coverage spans the directive split logic and the interceptor walk. Two concerns: (1) Must-fix data-loss bug in blocks.js when the BE pre-populates an answer with a raw id (no enrichment) — the page-load validation calls search with an undefined name and wipes the saved value; (2) related: campusName is not cleared when validation clears answer.value, causing UI/state desync. Tests don't yet cover the raw-id pre-populate scenario. Suggest extracting 'CAMPUS_V2' literals to constants across five call sites for typo safety.

app/scripts/directives/blocks/blocks.js
```diff
@@ -287,12 +296,37 @@ angular
...
-            if (!data.length) {
-              $scope.answer.value = '';
-            }
-          });
+          if ($scope.block.profileType === 'CAMPUS_V2') {
+            $scope.searchCampusesV2($scope.campusName).then((data) => {
+              if (!data.length) {
+                $scope.answer.value = '';
+              }
+            });
```
kegrimes: [Must-fix] When BE pre-populates with a raw id (no enrichment), campusName is undefined and this validation fires searchCampusesV2(undefined), returns empty, and wipes answer.value — data loss. Could we guard on `campusName` and also clear it on failure? [Category: File]

app/scripts/directives/blocks/blocks.js
```diff
@@ -287,12 +296,37 @@ angular
...
         if ($scope.answer.value) {
-          $scope.searchCampuses($scope.answer.value).then((data) => {
-            if (!data.length) {
-              $scope.answer.value = '';
-            }
-          });
+          if ($scope.block.profileType === 'CAMPUS_V2') {
+            $scope.searchCampusesV2($scope.campusName).then((data) => {
+              if (!data.length) {
+                $scope.answer.value = '';
```
kegrimes: [Must-fix] Clearing answer.value but leaving campusName leaves the input displaying a stale name with no underlying value. Could we add `$scope.campusName = '';` here too? [Category: File]

app/scripts/directives/blocks/blocks.js
```diff
@@ -267,13 +267,22 @@ angular
...
 angular
   .module('confRegistrationWebApp')
   .directive('campusQuestion', function () {
     return {
       templateUrl: campusQuestionTemplate,
       restrict: 'E',
       controller: function ($scope, $http) {
+        if (
+          $scope.block.profileType === 'CAMPUS_V2' &&
+          angular.isObject($scope.answer.value)
```
kegrimes: [Nice-to-have] Defensive on $scope.answer: `$scope.answer && angular.isObject($scope.answer.value)`. Matches the safety style used in the interceptor and avoids a TypeError if answer is unscaffolded. [Category: File]

app/scripts/directives/blocks/blocks.js
```diff
@@ -267,13 +267,22 @@ angular
...
+// All future campus questions should use CAMPUS_V2, but we need to keep the old one around for existing questions that use it
 angular
   .module('confRegistrationWebApp')
   .directive('campusQuestion', function () {
     return {
       templateUrl: campusQuestionTemplate,
       restrict: 'E',
       controller: function ($scope, $http) {
+        if (
+          $scope.block.profileType === 'CAMPUS_V2' &&
```
kegrimes: [Pattern] 'CAMPUS_V2' literal appears here, line 317, blockEditor.js:225, questionToolbar.js:175. Could we extract to a constant in services/constants.js? Prevents typo drift. [Category: Pattern]

app/scripts/directives/blocks/blocks.js
```diff
@@ -287,12 +296,37 @@ angular
...
+            name: val,
+          };
+          return $http
+            .get('campuses/connections/search', { params: $scope.params })
+            .then(function (campusNames) {
+              return campusNames.data.records;
+            });
+        };
+
+        $scope.selectCampus = function (campus) {
```
kegrimes, line 326: [Suggested tests] Add a spec that sets `answer.value = 'campus-id-string'` (raw id, no enrichment) and asserts no API call wipes it. This would have caught the data-loss bug at line 318-321. [Category: File]

app/scripts/directives/blocks/blocks.js
```diff
@@ -287,12 +296,37 @@ angular
...
+          $scope.answer.value = campus.id;
+          $scope.campusName = campus.name;
```
kegrimes, line 328: [Mentoring] Whenever you clear one field on a validation failure, scan for paired fields: here `answer.value` and `campusName` go together. Pair-state bugs cause ghost UI states where the input shows data the model doesn't have. [Category: File]

app/scripts/directives/blocks/blocks.spec.js
```diff
@@ -426,6 +443,50 @@ describe('Directive: blocks', () => {
 
       expect($scope.answer.value).toBe('');
     });
+
```
kegrimes, line 464: [Suggested tests] Both V2 tests use the enriched `{id,name}` shape. Add a case with `answer.value = '123'` (raw id from BE pre-populate) to guard against the data-loss bug in blocks.js:318. [Category: File]

app/scripts/directives/blocks/campusQuestion.html
```diff
@@ -19,6 +20,28 @@
     typeahead-no-results="campusNotFound"
     typeahead-show-hint="true"
   />
+  <input
+    ng-if="block.profileType === 'CAMPUS_V2'"
```
kegrimes: [Nice-to-have] Both V1 and V2 inputs share `id="block-input-{{block.id}}"`. ng-if prevents both rendering at once — just be careful if any selector/label depends on the id being unique per profile type. [Category: File]

app/scripts/directives/blocks/campusQuestion.html
```diff
@@ -19,6 +20,28 @@
...
+    show-errors
+    ng-required="block.required"
+    uib-typeahead="campus.name for campus in searchCampusesV2($viewValue)"
+    typeahead-on-select="selectCampus($item)"
+    typeahead-loading="loadingCampuses"
+    typeahead-min-length="2"
+    typeahead-wait-ms="400"
+    typeahead-editable="false"
+    typeahead-no-results="campusNotFound"
+    typeahead-show-hint="true"
```
kegrimes: [Mentoring] V2 ng-model points to `campusName` (local display state) while the typeahead's on-select writes the id to `answer.value` (persisted state). Splitting display from persisted state is a clean pattern — worth reusing. [Category: File]

app/scripts/directives/showAnswer/answerDisplay.html
```diff
@@ -37,4 +37,7 @@
   <span ng-switch-when="birthDateQuestion">
     {{answer.value | date:'MMMM dd, yyyy'}}</span
   >
+  <span ng-switch-when="campusQuestion">
+    {{answer.value.name || answer.value}}</span
+  >
```
kegrimes: [Nice-to-have] After a save, the interceptor mutates value to a string id in-place; until the next GET re-enriches, this renders the raw UUID via the fallback. Acceptable today, but if any flow lingers between save and re-fetch, deep-clone in the interceptor. [Category: File]

app/scripts/services/transformCampusAnswerInterceptor.js
```diff
@@ -0,0 +1,44 @@
+angular
+  .module('confRegistrationWebApp')
+  .factory('transformCampusAnswerInterceptor', function () {
+    // Transform campus answer objects to just their id before sending to the server
+    function transform(answer) {
```
kegrimes, line 5: [Nice-to-have] `answer &&` in transform is redundant given walk's gate at the call site (`item.blockId && item.value !== undefined`). Harmless defensive style — fine to keep. [Category: File]

app/scripts/services/transformCampusAnswerInterceptor.js
```diff
@@ -0,0 +1,44 @@
...
+        if (angular.isArray(item[key])) {
+          item[key].forEach(walk);
+        }
+      });
+    }
+
+    return {
+      request: function (config) {
+        // Only transform data on PUT and POST requests
+        if (
```
kegrimes, line 34: [Nice-to-have] Only PUT/POST handled. PATCH isn't used today, but if introduced later it would slip past. Worth a one-line comment locking in the rationale. [Category: File]

app/scripts/services/transformCampusAnswerInterceptor.js
```diff
@@ -0,0 +1,44 @@
...
+      // Handles different data shapes where campus answers may be nested
+      ['answers', 'registrants', 'groupRegistrants'].forEach(function (key) {
+        if (angular.isArray(item[key])) {
+          item[key].forEach(walk);
+        }
+      });
+    }
+
+    return {
+      request: function (config) {
```
kegrimes, line 32: [Mentoring] Interceptor mutates `config.data` in place — after a PUT fires, in-memory answer.value is a string. Other readers see the post-mutation shape until the next GET. Pairs well with BE enrichment-on-read; just know the seam. [Category: File]

app/scripts/services/transformCampusAnswerInterceptor.spec.js
```diff
@@ -0,0 +1,95 @@
...
+describe('Service: transformCampusAnswerInterceptor', function () {
+  beforeEach(angular.mock.module('confRegistrationWebApp'));
+
+  var transformCampusAnswerInterceptor;
+  beforeEach(inject(function (_transformCampusAnswerInterceptor_) {
+    transformCampusAnswerInterceptor = _transformCampusAnswerInterceptor_;
+  }));
+
+  it('transforms a single campus answer body on PUT', function () {
+    var config = {
```
kegrimes: [Suggested tests] Add one POST case mirroring the single-answer PUT test — cheap completeness check covering the second method branch. [Category: File]

app/scripts/directives/blockEditor/blockEditor.js
```diff
@@ -222,7 +222,7 @@ angular.module('confRegistrationWebApp').directive('blockEditor', function () {
       $scope.typeToProfile.yearInSchoolQuestion = 'YEAR_IN_SCHOOL';
       $scope.typeToProfile.opportunitiesQuestion = 'OPPORTUNITIES';
       $scope.typeToProfile.birthDateQuestion = 'BIRTH_DATE';
-      $scope.typeToProfile.campusQuestion = 'CAMPUS';
+      $scope.typeToProfile.campusQuestion = 'CAMPUS_V2';
```
kegrimes: [Nice-to-have] One-line comment mirroring blocks.js:270 ("All future campus questions should use CAMPUS_V2…") helps future maintainers understand why V2 over V1. [Category: File]

kegrimes: [Pattern] Hardcoded 'CAMPUS_V2' — same literal at 4 other sites. Could we centralize in services/constants.js? [Category: Pattern]

app/scripts/directives/questionToolbar/questionToolbar.js
```diff
@@ -172,7 +172,7 @@ angular
             id: 'campusQuestion',
             defaultTitle:
               'Campus - Type the full name of your campus and choose from the list that appears',
-            defaultProfile: 'CAMPUS',
+            defaultProfile: 'CAMPUS_V2',
```
kegrimes, line 175: [Pattern] Hardcoded 'CAMPUS_V2' literal duplicates with blocks.js and blockEditor.js. Centralize as a constant. [Category: Pattern]

app/scripts/directives/blocks/campusQuestion.html
kegrimes: /dismiss the id will still be unique since we are choosing only one question

app/scripts/directives/blockEditor/blockEditor.js
kegrimes: /dismiss not necessary

app/scripts/directives/blocks/blocks.js
kegrimes: /dismiss not necessary

app/scripts/directives/questionToolbar/questionToolbar.js
kegrimes: /dismiss not necessary

app/scripts/directives/blocks/campusQuestion.html
```diff
@@ -19,6 +20,28 @@
     typeahead-no-results="campusNotFound"
     typeahead-show-hint="true"
   />
+  <input
```
canac: Instead of branching here and duplicating all of the attributes, did you try checking `block.profileType` inside of inside `searchCampuses` and just branching in the `uib-typeahead` and `typeahead-on-select` attributes (and any others that are different in v2)? No worries if you tried that already and decided to go with this approach.

app/scripts/directives/blocks/blocks.js
```diff
@@ -287,12 +297,38 @@ angular
               return campusNames.data;
             });
         };
+
+        $scope.searchCampusesV2 = function (val) {
```
canac: I see `showInternationalCampuses` in the other campus searcher. Do we need to hide that checkbox in the UI for v2 campuses?

app/scripts/directives/blocks/blocks.js
```diff
@@ -267,13 +267,23 @@ angular
     };
   });
 
+// All future campus questions should use CAMPUS_V2, but we need to keep the old one around for existing questions that use it
```
canac: praise: Thanks for keeping support for the old campus type!

canac (commented):
This looks really good! In the admin Registrations view, the campus just shows its ID until I expand it. Is it possible to fix that?

Image

Image

app/scripts/directives/blocks/campusQuestion.html
kegrimes: I was thinking it might be easier to delete the first campus question once all the old events are done using it! But if it would be better to do this, I can totally change it! What do you think?

app/scripts/directives/blocks/blocks.js
kegrimes: This should only be an optional parameter for the original search. I wasn't aware there was a checkbox, could you send a screenshot of where you see it in the v2 question?

canac: If you edit the question in the admin view, you can see that checkbox.

Screenshot 2026-05-20 at 7 50 01 AM

app/scripts/directives/blocks/campusQuestion.html
canac: > I was thinking it might be easier to delete the first campus question once all the old events are done using it!

You should see what Ben things, but a lot of times annual events clone old ERT events indefinitely. So unless we can make v1 campuses upgrade to v2 campuses during cloning, we're going to have v1 campus questions hanging around for a very long time.

canac:
If you haven't yet, I'd also check that the campus name shows up instead of the ID in a CSV download from the admin registrations page. Although, I think the CSV is generated server-side, so it's on Ben to get that working correctly.

app/scripts/directives/blocks/blocks.js
kegrimes: Thanks for finding this! I will get on removing that.

app/scripts/directives/blocks/campusQuestion.html
kegrimes: Ben said he could completely replace CAMPUS with CAMPUS_V2, however, cloning is already a fragile enough process and he is sure some edge cases would appear. He is more for letting the user's update the question themselves. He added a 400 response to tell users to remove the school question before cloning and they can re-add it before or after. This should be okay because the question will be re-added as CAMPUS_V2!

kegrimes:
@canac Most of the issues you flagged were backend issues! Ben fixed the id showing up instead of the campus name in the admin view, however, there is still one tiny tweak he is making for the exports which should work after his changes! 

I removed the checkbox, so everything else should be good now! I just wanted to catch you before you are off for a week.

Also if his PR is only in staging, I should wait to merge this correct? (https://github.com/CruGlobal/ert-api/pull/1092)

canac (approved):
It works great! Yes, this PR shouldn't merge until Ben's PR does. And when you do merge, try to do it early in the week like Monday-Wednesday.

kegrimes:
@canac I misunderstood that there was a new block type, `campusV2Question`, alongside the new profile type `CAMPUS_V2`. It would be an easier change on the frontend, so I reworked it. Can you take a tiny look to make sure everything is good!

kegrimes:
@zweatshirt Could you glance over this for me? Caleb already approved the changes but I had to rework some things on Friday! I forgot he was out all week.

app/scripts/services/transformCampusAnswerInterceptor.js
```diff
@@ -0,0 +1,43 @@
...
+        // Only transform data on PUT and POST requests
+        if (
+          (config.method === 'PUT' || config.method === 'POST') &&
+          config.data
```
zweatshirt: Question: I'm wondering if we can narrow this to just the registrations url, to avoid calling the walk function on every PUT or POST, which may be unrelated to registration. What do you think? It's possible we can perform this flattening without the need for an interceptor, but that may not be a trivial change.

app/scripts/services/transformCampusAnswerInterceptor.spec.js
```diff
@@ -0,0 +1,95 @@
+import 'angular-mocks';
+
+describe('Service: transformCampusAnswerInterceptor', function () {
+  beforeEach(angular.mock.module('confRegistrationWebApp'));
+
+  var transformCampusAnswerInterceptor;
```
zweatshirt: Question: Do we want to use `let` and `const` in this file? I think we have been trying to get away from using `var`

zweatshirt (approved):
Seems to work well from what I can tell. Looks great Katelyn!!

app/scripts/services/transformCampusAnswerInterceptor.js
kegrimes: This is a good point. I think we could do something like this where the data is found.

```
/\/(answers|registrants|registrations)(\/|\?|$)/.test(config.url)
```

app/scripts/services/transformCampusAnswerInterceptor.spec.js
kegrimes: Yes, thanks for catching that! I had Claude write most of these tests

Benjamin-deVries:
@kegrimes The API just deployed.

PR #983, opened 2026-05-01
EVENT-1001 - Maintenance: Increase test times

zweatshirt:
## Description

[Jira ticket](https://jira.cru.org/browse/EVENT-1001)
ERT maintenance.
Next step would be to consider Jest, which is a larger refactor. This is a quick win.

### Changes

- **[webpack.config.js](webpack.config.js)**: Added a webpack filesystem cache scoped to test mode, with the cache version keyed by whether coverage is enabled.
- **[babel-angular.config.js](babel-angular.config.js)**: Made `babel-plugin-istanbul` (coverage instrumentation) conditional on `CI=true` or `COVERAGE=true` instead of running on every test invocation.
- **[webpack.config.js](webpack.config.js)**: Skipped `ESLintPlugin` when building the test bundle.
- **[package.json](package.json)**: Added `yarn test:coverage` as the explicit local opt-in for a coverage report.

## Test time improvements

Measuring full `yarn test` (karma + jest) on master vs this branch.

| Scenario | master | this branch | Improvement |
|---|---|---|---|
| `yarn test` (cold) | 11.5s | 11.9s | ~same |
| `yarn test` (warm) | 10.9s | **7.3s** | **−33%** |
| Karma webpack compile (cold) | 3.85s | 2.96s | −23% |
| Karma webpack compile (warm) | 3.71s | **0.22s** | **−94%** |

### Where the win lands

The win is on **repeated local runs**. Once webpack's filesystem cache is populated, the karma webpack compile drops from ~3.7s to ~0.2s — that accounts for the entire 3.6s improvement.

Cold runs are roughly unchanged: the cache is empty on first run, so it can't help. Skipping ESLint and istanbul in test mode does shave ~0.9s off of most cold runs, but run-to-run variance puts both near 11.5s.

Cove coverage reports from instanbul are now opt-in: `yarn test:coverage` instead of the default behavior.

### CI impact

CI runs are unaffected. GitHub Actions runners are ephemeral, so the cache never has a hit; coverage is still generated (`CI=true` keeps istanbul on); codecov uploads continue to work.

## Testing

- Run tests on master and tests on this branch. Compare run time. The warm runs should be noticeably faster.

## Checklist

- [ ] I have requested a review from another person on the project

### DIFF
--- a/babel-angular.config.js
+++ b/babel-angular.config.js
@@ -23,8 +23,11 @@ module.exports = {
   env: {
     test: {
       plugins: [
-        'istanbul',
-        // Must run AFTER istanbul to properly annotate DI
+        ...(process.env.CI === 'true' || process.env.COVERAGE === 'true'
+          ? ['istanbul']
+          : []),
+        // Order-sensitive: when istanbul is included, angularjs-annotate must
+        // come after it to annotate the instrumented DI.
         'angularjs-annotate',
       ],
     },
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -13,6 +13,7 @@ const ESLintPlugin = require('eslint-webpack-plugin');
 
 const isBuild = (process.env.npm_lifecycle_event || '').startsWith('build');
 const ci = process.env.CI === 'true';
+const coverage = ci || process.env.COVERAGE === 'true';
 const prod = process.env.GITHUB_REF === 'refs/heads/master';
 
 const htmlMinDefaults = {
@@ -34,6 +35,17 @@ module.exports = (env = {}) => {
   const isTest = env.test;
   return {
     mode: isBuild ? 'production' : 'development',
+    // Cache compiled modules to avoid cold webpack rebuild during retesting.
+    cache: isTest
+      ? {
+          type: 'filesystem',
+          buildDependencies: { config: [__filename] },
+          // Babel emits instrumented vs. uninstrumented modules depending on
+          // whether coverage is on (see babel-angular.config.js), so each
+          // value gets its own cache slot.
+          version: `coverage=${coverage}`,
+        }
+      : undefined,
     entry: {
       app: ['scripts/main.js', 'styles/style.scss'],
     },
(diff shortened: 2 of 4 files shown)


webpack.config.js
```diff
@@ -72,14 +84,19 @@ module.exports = (env = {}) => {
...
         }),
-        new ESLintPlugin({
-          extensions: ['js', 'ts', 'tsx'],
-
-          // Show errors as warnings during development to prevent start/test commands from exiting
-          failOnError: isBuild || ci,
-          emitWarning: !isBuild && !ci,
-        }),
       ],
+      // eslint is its own job, we can skip it during tests
```
kegrimes: Great find! Definitely better to keep it separate from the tests

babel-angular.config.js
```diff
@@ -23,8 +23,11 @@ module.exports = {
   env: {
     test: {
       plugins: [
-        'istanbul',
-        // Must run AFTER istanbul to properly annotate DI
+        ...(process.env.CI === 'true' || process.env.COVERAGE === 'true'
+          ? ['istanbul']
+          : []),
```
kegrimes, line 28: Should the `coverage-istanbul` reporter in `karma.conf.js` also be made conditional? I think the reporter still runs on every `yarn test`, so there's nothing for it to report.

kegrimes (approved):
Really nice work Zach! 

I'm not sure how confident I feel reviewing this, so if you want to double check with Caleb that would be good, but what I'm seeing looks good.

babel-angular.config.js
zweatshirt: I think you're right, good catch

webpack.config.js
```diff
@@ -72,14 +84,19 @@ module.exports = (env = {}) => {
...
+      !isTest
```
canac, line 90: Should we merge this with the other `!isTest` block?

canac (approved):
I'm not a webpack wizard, so I don't have any concerns with this. If it build correctly in staging, I'm good! I assume the caching rebuilds files when they change? So we don't get stale test results? If you haven't verified this, it would great to by editing a test or source file with a cache present and checking that tests still fail.

I love that you found some easy wins to speed up the dev server and testing workflow!

zweatshirt:
> I'm not a webpack wizard, so I don't have any concerns with this. If it build correctly in staging, I'm good! I assume the caching rebuilds files when they change? So we don't get stale test results? If you haven't verified this, it would great to by editing a test or source file with a cache present and checking that tests still fail.
> 
> I love that you found some easy wins to speed up the dev server and testing workflow!

Yeah the cache is supposed to update. I believe I did a similar test to what you're suggesting but I'll also double check, never hurts to look again. The one concern, apparently (according to Claude), is that users have reported adding new test files sometimes gets missed by the cache, but it's rare (*apparently*...). I'll test that too

PR #984, opened 2026-05-06
EVENT-1002 - Reorder project: colocate views and scripts

zweatshirt:
## Description

Change project structure to be like this:
```
app/
├── features/
│   ├── help/
│   │   ├── help.controller.js
│   │   └── help.html
```

For readability, controllers were given a `controller` suffix.

I considered further making `/views` and `/scripts` subdirectories in each feature directory, but this seems sufficient for now.

[Jira ticket](https://jira.cru.org/browse/EVENT-1002)

## Testing

- It is extremely important to ensure that the application still behaves as expected, with no new errors or views that fail to resolve.
- Run through the application and test key features to ensure they still behave identically to before.

## Checklist


zweatshirt:
@canac I was really just hoping to get an opinion on this new structure, and a user test. It's good to have second eyes for a change like this. If you happen to catch anything odd would you let me know? Thanks for considering! For now, I'm going to avoid putting this on staging until I condense the commits, but preview works fine.

canac (commented):
This is a great start! If possible, I'd love to see directive templates in the same directory as the directive script. Is it possible to get rid of the views/ directory entirely? So all of the HTML files are either in a directive or feature subdirectory?

I was thinking that all of the tests in test/spec/ would be moved next to the code that they test. Did you run into issues trying to do that?

zweatshirt:
> This is a great start! If possible, I'd love to see directive templates in the same directory as the directive script. Is it possible to get rid of the views/ directory entirely? So all of the HTML files are either in a directive or feature subdirectory?
> 
> I was thinking that all of the tests in test/spec/ would be moved next to the code that they test. Did you run into issues trying to do that?

Sounds like a good idea, I'll do that. I don't think I ran into any issues moving those tests, I was just unsure to what extent everything should be moved. But makes sense if we commit to this project structure I might as well go all the way with it.

zweatshirt:
Update: I'll probably wait to merge this early next week, as Charles is generally against merging large changes close to the weekend and we are active with FL.

app/scripts/directives/blocks/blocks.spec.js
canac, line 1: Want to pluralize this to `blocks.spec.js` to match the JS file?

app/scripts/services/httpUrlInterceptor.js
```diff
@@ -3,11 +3,7 @@ angular
   .factory('httpUrlInterceptor', function (envService) {
     return {
       request: function (config) {
-        var passthroughRegexs = [
-          /https?:\/\/.*/,
-          /^\/views\/.*/,
-          /template\/.*/,
-        ];
+        var passthroughRegexs = [/https?:\/\/.*/, /\.html$/];
```
canac, line 6: praise: I like this a lot better than whitelisting directories! Now our HTML files can live anywhere.

jest.config.js
```diff
@@ -10,11 +10,13 @@ const config = {
...
   ],
   coverageDirectory: 'coverage/jest',
-  moduleDirectories: ['node_modules', 'views'],
+  moduleDirectories: ['node_modules'],
   moduleNameMapper: {
     '\\.(html)$': '<rootDir>/__tests__/templateMock.ts',
   },
   roots: ['<rootDir>/app/scripts/'],
+  testMatch: ['**/__tests__/**/*.[jt]s?(x)', '**/*.test.[jt]s?(x)'],
+  testPathIgnorePatterns: ['/node_modules/', '\\.spec\\.js$'],
```
canac: question: Does this work? I know we have files in `__tests__`, but they are helpers, not actual tests.

```suggestion
  testMatch: ['**/*.test.[jt]s?(x)'],
```

canac (approved):
This is amazing! It will be a lot easier to navigate the project.

zweatshirt:
Staging bot isn't working

PR #985, opened 2026-05-06
No-Jira - Fix Netlify staging/development inconsistency

zweatshirt:
## Description

Netlify migrated deploy-preview hostnames from *.netlify.com to *.netlify.app years ago, but our envService domain list still pointed at the stale .com pattern. As a result, deploy previews fell through to the development env which happened to work only because development and staging share the same staging API URL.

### This PR:

- Adds an explicit preview environment matching *.netlify.app, using the staging API + TSYS sandbox (so payment flows tested on previews never hit live processing).
- Removes the now-stale *.netlify.com entry from staging. Any URL still on the old domain (extremely unlikely) would fall through to development.
- Extends the $compileProvider.debugInfoEnabled(false) gate to include preview, turning off Angular's debug-info classes/methods on deploy previews — small perf win on data-heavy pages.
- Changes ERT's staging and preview TSYS environments to staging instead of production.

Question: Do we want to instead just re-add Netlify correctly to the staging config?

## Testing

- Test preview environment behavior: https://deploy-preview-985--eventregistrationtool.netlify.app/

## Checklist

### DIFF
--- a/app/scripts/app.config.js
+++ b/app/scripts/app.config.js
@@ -38,7 +38,8 @@ angular
     envServiceProvider.config({
       domains: {
         development: ['localhost'],
-        staging: ['stage.eventregistrationtool.com', '*.netlify.com'],
+        preview: ['*.netlify.app'],
+        staging: ['stage.eventregistrationtool.com'],
         production: [
           'www.eventregistrationtool.com',
           'eventregistrationtool.com',
@@ -46,6 +47,11 @@ angular
       },
       vars: {
         development: {
+          apiUrl:
+            'https://api.stage.eventregistrationtool.com/eventhub-api/rest/',
+          tsysEnvironment: 'staging',
+        },
+        preview: {
           apiUrl:
             'https://api.stage.eventregistrationtool.com/eventhub-api/rest/',
           tsysEnvironment: 'staging',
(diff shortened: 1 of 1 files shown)


kegrimes (approved):
This looks great! And to answer your question, I think it is good to keep `preview` separate. If we were to fold it into staging, I believe payment testing would go through production TSYS environment. 

```
staging: {
  apiUrl:
    'https://api.stage.eventregistrationtool.com/eventhub-api/rest/',
  tsysEnvironment: 'production',
},
```

I opened the preview URL and ERT seems to behave normally. I didn't dig too far into payment flows, but this looks correct.

zweatshirt:
> This looks great! And to answer your question, I think it is good to keep `preview` separate. If we were to fold it into staging, I believe payment testing would go through production TSYS environment.
> 
> ```
> staging: {
>   apiUrl:
>     'https://api.stage.eventregistrationtool.com/eventhub-api/rest/',
>   tsysEnvironment: 'production',
> },
> ```
> 
> I opened the preview URL and ERT seems to behave normally. I didn't dig too far into payment flows, but this looks correct.

Yeah, TSYS would be prod, but I'm not sure if that's actually a bad thing or not. If it is prod in staging there is an argument it could/should be prod in preview environments (since in most cases these environments should behave identically). It actually makes me wonder as well if ERT staging's TSYS environment should be staging instead of production.

zweatshirt:
@kegrimes 

> This looks great! And to answer your question, I think it is good to keep `preview` separate. If we were to fold it into staging, I believe payment testing would go through production TSYS environment.

I asked Ben about it, I think we'll want to change ERT staging to point to TSYS staging as well. Not sure why it was set to production 🤔

PR #987, opened 2026-05-12
[No-Jira] Disable account number during registration

kegrimes:
## Description

During SIT2 testing, it was brought up that we want to disable the "Account Number" field on Staff transfers during registration.

Things to note:
- We want to ensure we do not disable the field when `employeeId` is being used to preserve the original logic (not 100% sure if this is needed, but I don't think it's a bad idea)
- We also don't want to disable this field when an admin is making the payment for a registrant

## Testing

Registration Test: with staff account id ✅
- Go to `/register/3b88f5e9-545f-4cb8-8640-1bb252acf0ff/page/`
- Register for "Ben's Test Event"
- During payment, select "Cru Staff / Ministry Account Transfer" and "Staff"
- Check that the account number field is disabled with correct staff account id

Admin payment test: ✅
- Go to a registrant in your own event
- Click the green button under "Paid"
- Go to "Add" tab and select a staff transfer
- Verify the account number field is enabled with nothing populated

Registration Test: without staff account id ✅
Screenshot 2026-05-12 at 4 01 50 PM

## Checklist

### DIFF
--- a/app/scripts/directives/payment.js
+++ b/app/scripts/directives/payment.js
@@ -36,6 +36,7 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function () {
       $scope.currentYear = new Date().getFullYear();
       $scope.creditCardCountry = 'US';
       $scope.countries = allCountries;
+      $scope.disable = false;
 
       $scope.paymentMethodsViews = {
         CREDIT_CARD: creditCardTemplate,
@@ -382,6 +383,8 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function () {
               profile.staffAccountNumber ||
               transformEmployeeIdIntoAccountNumber() ||
               '';
+
+            $scope.disable = !!profile.staffAccountNumber;
           },
           function () {
             $scope.currentPayment.transfer.accountNumber = '';
@@ -397,6 +400,7 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function () {
           fetchStaffAccountNumber();
         } else {
           $scope.currentPayment.transfer.accountNumber = '';
+          $scope.disable = false;
         }
         [
           $scope.currentPayment.transfer.businessUnit,
--- a/app/scripts/directives/payment.js
+++ b/app/scripts/directives/payment.js
@@ -36,7 +36,7 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function () {
       $scope.currentYear = new Date().getFullYear();
       $scope.creditCardCountry = 'US';
       $scope.countries = allCountries;
-      $scope.disable = false;
+      $scope.accountNumberDisabled = false;
 
       $scope.paymentMethodsViews = {
         CREDIT_CARD: creditCardTemplate,
@@ -373,6 +373,8 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function () {
       }
 
       function fetchStaffAccountNumber() {
+        $scope.accountNumberDisabled = false;
+
         // staffAccountNumber is fetched asynchronously after login
         // and may not be in the cached profile yet, so refetch to pick it up
         ProfileCache.clearCache();
@@ -384,10 +386,11 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function () {
               transformEmployeeIdIntoAccountNumber() ||
               '';
 
-            $scope.disable = !!profile.staffAccountNumber;
+            $scope.accountNumberDisabled = !!profile.staffAccountNumber;
           },
           function () {
             $scope.currentPayment.transfer.accountNumber = '';
+            $scope.accountNumberDisabled = false;
           },
         );
       }
@@ -400,7 +403,7 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function () {
           fetchStaffAccountNumber();
         } else {
           $scope.currentPayment.transfer.accountNumber = '';
-          $scope.disable = false;
+          $scope.accountNumberDisabled = false;
         }
         [
           $scope.currentPayment.transfer.businessUnit,
--- a/app/views/paymentMethods/transfer.html
+++ b/app/views/paymentMethods/transfer.html
@@ -84,6 +84,7 @@
         ng-model="currentPayment.transfer.accountNumber"
         type="text"
         class="form-control"
+        ng-disabled="disable"
       />
     </label>
   </div>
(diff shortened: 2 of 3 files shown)


app/scripts/directives/payment.js
```diff
@@ -382,6 +383,8 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function () {
               profile.staffAccountNumber ||
               transformEmployeeIdIntoAccountNumber() ||
               '';
+
+            $scope.disable = !!profile.staffAccountNumber;
```
kegrimes: [Must-fix] The rejection callback at lines 389-391 clears `accountNumber` but doesn't reset `$scope.disable`. If a prior success set it true and a later fetch rejects, the field ends up empty AND locked — user can't type. Could we add `$scope.disable = false;` to the error branch? [Category: File]

kegrimes: [Must-fix] Async race: if the user toggles STAFF → other → STAFF quickly, an older `ProfileCache.getCache()` may resolve after a newer one, leaving `disable` stale. Could we reset `$scope.disable = false;` at the top of `fetchStaffAccountNumber` (just before the `ProfileCache.getCache()` call) so the new fetch always starts from a clean state? [Category: File]

kegrimes (commented):
Small, focused change introducing `$scope.disable` to lock the staff Account Number input when HCM returns an authoritative value. Two correctness issues worth addressing before merge: the rejection branch of `fetchStaffAccountNumber` doesn't reset `disable` (locks the user out on error recovery), and async toggles can leave `disable` stale if the user changes account types faster than the profile fetch resolves. Flag name `disable` is too generic on a shared directive scope — `accountNumberDisabled` will age better. Tests cover the happy paths well but should also assert disable resets in the rejection spec and exercise the employeeId digit-strip transform with a non-numeric input. No reuse, i18n, or DOM-manipulation concerns.

app/scripts/directives/payment.js
```diff
@@ -36,6 +36,7 @@ angular.module('confRegistrationWebApp').directive('ertPayment', function () {
       $scope.currentYear = new Date().getFullYear();
       $scope.creditCardCountry = 'US';
       $scope.countries = allCountries;
+      $scope.disable = false;
```
kegrimes: [Nice-to-have] `$scope.disable` is generic on a shared directive scope (sibling props: `currentPayment`, `accountTypeChanged`, `creditCardCountry`). Could we rename to `accountNumberDisabled` so future ng-disabled bindings here don't collide? [Category: File]

app/scripts/directives/payment.js
kegrimes: [Mentoring] Idiomatic AngularJS — flag + `ng-disabled` is the right shape. Two habits to lean into: defensively reset transient state in every exit path of an async fn, and name `$scope`-shared flags for their concern, not their action. [Category: File]

kegrimes: [Mentoring] TODO at L381 says the employeeId fallback goes away once HCM ships. After that, `!!profile.staffAccountNumber` is essentially always true for non-admin STAFF and the lock becomes always-on. A one-line comment here would document why the flag exists. [Category: File]

kegrimes: [Suggested tests] Add a spec asserting `scope.disable === false` after `ProfileCache.getCache()` rejects, and a rapid-toggle spec covering STAFF → not-STAFF → STAFF-with-rejection ending in `disable === false`. [Category: File]

app/views/paymentMethods/transfer.html
```diff
@@ -84,6 +84,7 @@
         ng-model="currentPayment.transfer.accountNumber"
         type="text"
         class="form-control"
+        ng-disabled="disable"
```
kegrimes: [Nice-to-have] If we rename the flag to `accountNumberDisabled`, this binding becomes self-documenting: `ng-disabled="accountNumberDisabled"`. Pure readability. [Category: File]

test/spec/directives/payment.spec.js
```diff
@@ -58,6 +58,24 @@ describe('Directive: ertPayment', function () {
...
+    );
+    scope.currentPayment = {
+      transfer: { accountType: 'STAFF', accountNumber: '123' },
+    };
+    scope.accountTypeChanged();
+    scope.$apply();
+
+    expect(scope.currentPayment.transfer.accountNumber).toBe('1234567');
+    expect(scope.disable).toBe(false);
   });
```
kegrimes, line 80: [Nice-to-have] Test name says "prefill employeeId" but the directive transforms it (digits-only, last-7). Mock `'1234567'` is already 7 digits so the transform is invisible. Could we add a variant with non-digits (e.g., `'EMP-1234567A'`) to cover the regex? [Category: File]

test/spec/directives/payment.spec.js
```diff
@@ -58,6 +58,24 @@ describe('Directive: ertPayment', function () {
     scope.$apply();
 
     expect(scope.currentPayment.transfer.accountNumber).toBe('9870123457');
+    expect(scope.disable).toBe(true);
```
kegrimes: [Suggested tests] The existing rejection spec at L42-51 should also assert `expect(scope.disable).toBe(false);` once the rejection-branch fix lands, to lock the behavior in. [Category: File]

app/scripts/directives/payment.js
kegrimes: /dismiss not necessary

kegrimes: /dismiss not necessary, coverage is met

kegrimes:
@canac Could you test this for me? During SIT2 testing this morning, there was discussion around disabling the "Account Number" field during registration. I can't test since I don't have a staff account id, but I have already tested the `employeeId` case!

app/views/paymentMethods/transfer.html
```diff
@@ -84,6 +84,7 @@
         ng-model="currentPayment.transfer.accountNumber"
         type="text"
         class="form-control"
+        ng-disabled="accountNumberDisabled"
```
canac, line 87: optional: I just read an article about how readonly is more accessible than disabled https://zeroheight.com/blog/rethinking-the-disabled-norm/

canac (approved):
I can't test it myself, but the code looks great! You could try a DevTools network override to give your profile a staff account. Or wait for my Okta account to be working again, and I can retest.

app/views/paymentMethods/transfer.html
kegrimes: Hmm that's interesting! I have never thought about how disabling could affect accessibility. It makes me wonder if this would be a good thing in the future to look into so we can make our apps more accessible!

kegrimes: I will probably stick with disabled for now since it is consistent with the rest of ERT!

PR #989, opened 2026-06-10
[HS-1621020] Handle missing currency

canac:
## Description

Some conferences don't have a currency set. This can cause `localizedCurrency` throw an exception, preventing the page from rendering.

After I made this PR, I saw the user's response that they have a currency set. So this may not fix their bug. But it will at least fix it for other conferences and make it easier to debug future issues like this.

I also cleaned up the `localizedSymbol` implementation. We don't need feature checks for `Intl.NumberFormat.formatToParts` anymore.

HelpScout ticket: https://secure.helpscout.net/conversation/3326003234/1621020/

## Testing

Not directly testable. But I tested locally rendering `| localizedCurrency` in a conference without a currency. The tests also exercise this.

## Checklist

### DIFF
--- a/app/scripts/filters/localizedCurrency.js
+++ b/app/scripts/filters/localizedCurrency.js
@@ -12,19 +12,3 @@ angular
       });
     };
   });
-
-angular
-  .module('confRegistrationWebApp')
-  .filter('localizedSymbol', ($locale) => {
-    return (currencyCode) => {
-      const localeId = $locale && $locale.id ? $locale.id : 'en-us';
-      const numberFormat = new Intl.NumberFormat(localeId, {
-        style: 'currency',
-        currency: currencyCode || 'USD',
-      });
-      const currencySymbol = numberFormat
-        .formatToParts(0)
-        .find((part) => part.type === 'currency')?.value;
-      return currencySymbol;
-    };
-  });
--- a/app/scripts/filters/localizedCurrency.js
+++ b/app/scripts/filters/localizedCurrency.js
@@ -1,3 +1,5 @@
+const defaultCurrency = 'USD';
+
 angular
   .module('confRegistrationWebApp')
   .filter('localizedCurrency', function ($locale) {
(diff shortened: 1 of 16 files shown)


app/scripts/filters/localizedCurrency.js
```diff
@@ -2,51 +2,29 @@ angular
...
-        })
-        .replace(/[0., ]/g, '');
+      const localeId = $locale && $locale.id ? $locale.id : 'en-us';
+      const numberFormat = new Intl.NumberFormat(localeId, {
+        style: 'currency',
+        currency: currencyCode || 'USD',
+      });
+      const currencySymbol = numberFormat
+        .formatToParts(0)
+        .find((part) => part.type === 'currency')?.value;
```
zweatshirt: `.find((part) => part.type === 'currency')?.value;` 
Will this ever return an undefined value? 

Additionally, the 'USD' fallback would correctly handle false-y values, but if `currencyCode` were ever an invalid truth-y value I think `NumberFormat` would throw before we can ever set `currencySymbol`.

app/scripts/filters/localizedCurrency.spec.js
```diff
@@ -16,6 +16,20 @@ describe('Filter: localizedCurrency', function () {
...
   });
+
+  it('should format zero amounts', function () {
+    expect(filter(0, 'USD')).toContain('0.00');
+  });
+
+  it('should fall back to USD when currency code is missing', function () {
+    expect(filter(123.12, undefined)).toContain('123.12');
+    expect(filter(123.12, null)).toContain('123.12');
+  });
```
zweatshirt, line 27: Should we test the truthy but invalid currencyCode case I mentioned?

zweatshirt (approved):
I think this is likely good as is, as the concern I have about truth-y but invalid currencyCode values is probably unlikely and ideally a backend concern, so I am going to approve. I am interested to get your view on that though.

app/scripts/filters/localizedCurrency.js
canac: I don't think so, but the server provides us with a currency `shortSymbol`, so I removed `localizedSymbol` entirely. Now we rely on the server's currency short code.

PR #991, opened 2026-06-11
[HS-1621020] Stop throwing when a question references a rule that was deleted

canac:
## Description

A particular conference is having issues when people review their completed registrations to pay for it. This issue is likely a crash when `parentBlock` is undefined.

HelpScout ticket: https://secure.helpscout.net/conversation/3326003234/1621020/
Rollbar occurrence: https://app.rollbar.com/a/Cru/fix/item/ert-web/1017/occurrence/462407511837#detail

## Testing

- *I'm still working on recreating the issue locally and verifying the fix*

## Checklist

### DIFF
--- a/app/scripts/services/validateRegistrant.js
+++ b/app/scripts/services/validateRegistrant.js
@@ -82,10 +82,12 @@ angular
                 ? parseFloat(rule.value)
                 : rule.value;
             }
-            // Hide this block if the question that the rule is based on (parentBlock) is hidden.
+            // Hide this block if the question that the rule is based on (parentBlock) is hidden or deleted.
+            // The parentBlock may have been deleted, leaving an orphaned rule.
             let parentBlock = _.find(blocks, { id: rule.parentBlockId });
             if (
               conference &&
+              parentBlock &&
               !blockVisibleRuleCheck(
                 parentBlock,
                 registrant,
--- a/app/scripts/services/validateRegistrant.js
+++ b/app/scripts/services/validateRegistrant.js
@@ -82,12 +82,11 @@ angular
                 ? parseFloat(rule.value)
                 : rule.value;
             }
-            // Hide this block if the question that the rule is based on (parentBlock) is hidden or deleted.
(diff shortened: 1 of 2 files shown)


app/scripts/services/validateRegistrant.js
```diff
@@ -82,10 +82,12 @@ angular
...
                 : rule.value;
             }
-            // Hide this block if the question that the rule is based on (parentBlock) is hidden.
+            // Hide this block if the question that the rule is based on (parentBlock) is hidden or deleted.
+            // The parentBlock may have been deleted, leaving an orphaned rule.
             let parentBlock = _.find(blocks, { id: rule.parentBlockId });
             if (
               conference &&
+              parentBlock &&
               !blockVisibleRuleCheck(
```
zweatshirt: This file is kind of hard to reason about. I am wondering if the change we want here is actually `(!parentBlock || !blockVisibileRuleCheck(...))`?

canac: It really is 🤯. I did more investigation, and the issue is admin-only parent blocks being silently not returned by the server, not those parents being deleted. If the parent block is missing because it is admin-only, I think it should be treated the same way as hidden. Any questions dependent on it shouldn't show up. So I believe this is correct.

zweatshirt: Ahhh okay, that makes more sense.

app/scripts/services/validateRegistrant.js
```diff
@@ -86,6 +86,7 @@ angular
             let parentBlock = _.find(blocks, { id: rule.parentBlockId });
             if (
               conference &&
+              parentBlock && // the server strips out admin-only blocks, so the parent might be missing
```
zweatshirt, line 89: Thanks for adding this comment

zweatshirt (approved):
This seems right for the issue you mentioned, I am just concerned it could lead to separate bugs we aren't aware of. I'll trust your judgment!

PR #994, opened 2026-07-06
[No-Jira] Fix null profile types for CAMPUS_V2

kegrimes:
## Description

HelpScout ticket [#1665844](https://secure.helpscout.net/conversation/3373373152/1665844?viewId=8728581) raised issues with the connections school id displaying instead of the school name. I found that the profile type was `null` when it should have been `CAMPUS_V2` which was causing the bug.

Upon further investigation, I found out there are 3 ways a campus v2 question can have a `null` profile type:
1. Unchecking "Cru Profile" when editing the campus question
2. Copying the campus v2 question and then deleting the original
3. Dragging a duplicate campus v2 question and then deleting the original

To prevent this from happening, we need to restore the profile type on save to ensure no campus question persists without it.

**Note:** This will only fix events after an update is made, so I asked Ben if he can see how many events were affected and if he can change the profile type to `CAMPUS_V2`

## Testing

- Go to event dashboard and create an event with a campus question
- Go to registration page and ensure university appears (you must already have a connections id set)
- Go back to event and try # 1 above
- Navigate back to registration page and verify university is still there
- Repeat with # 2 and # 3

## Checklist

### DIFF
--- a/app/features/eventForm/eventForm.controller.js
+++ b/app/features/eventForm/eventForm.controller.js
@@ -45,7 +45,7 @@ angular
       var formSaving = false;
       var formSavingTimeout;
       var formSavingNotifyTimeout;
-      
+
       function normalizeCampusProfileTypes() {
         (
           ($scope.conference && $scope.conference.registrationPages) ||
--- a/app/features/eventForm/eventForm.controller.js
+++ b/app/features/eventForm/eventForm.controller.js
@@ -45,6 +45,22 @@ angular
       var formSaving = false;
       var formSavingTimeout;
       var formSavingNotifyTimeout;
+      
+      function normalizeCampusProfileTypes() {
+        (
+          ($scope.conference && $scope.conference.registrationPages) ||
+          []
+        ).forEach(function (page) {
+          (page.blocks || []).forEach(function (block) {
+            if (
+              block.type === 'campusV2Question' &&
+              block.profileType !== 'CAMPUS_V2'
+            ) {
+              block.profileType = 'CAMPUS_V2';
+            }
+          });
+        });
+      }
 
       function saveForm() {
         $timeout.cancel(formSavingTimeout);
@@ -56,6 +72,7 @@ angular
         }
 
         formSaving = true;
+        normalizeCampusProfileTypes();
         let conferenceWithoutImage = angular.copy($scope.conference);
         conferenceWithoutImage.image = null;
 
--- a/app/features/eventForm/eventForm.controller.spec.js
+++ b/app/features/eventForm/eventForm.controller.spec.js
@@ -138,9 +138,11 @@ describe('Controller: eventForm', function () {
       const page = scope.conference.registrationPages[0];
 
       scope.$apply(() => {
-        page.blocks.push(
-          { id: 'campus-v2-block', type: 'campusV2Question', profileType: null },
-        );
+        page.blocks.push({
+          id: 'campus-v2-block',
+          type: 'campusV2Question',
+          profileType: null,
+        });
       });
       $httpBackend.flush();
 
--- a/app/features/eventForm/eventForm.controller.spec.js
+++ b/app/features/eventForm/eventForm.controller.spec.js
@@ -127,6 +127,29 @@ describe('Controller: eventForm', function () {
     });
   });
 
+  describe('normalizeCampusProfileTypes', () => {
+    beforeEach(() => {
+      scope.$digest();
+    });
+
+    it('forces campusV2Question blocks to the CAMPUS_V2 profile type on save', () => {
+      spyOn(ConfCache, 'update');
+      $httpBackend.expectPUT(/^conferences\/.+$/).respond(204, '');
+      const page = scope.conference.registrationPages[0];
+
+      scope.$apply(() => {
+        page.blocks.push(
+          { id: 'campus-v2-block', type: 'campusV2Question', profileType: null },
+        );
+      });
+      $httpBackend.flush();
+
+      expect(
+        page.blocks.find((block) => block.id === 'campus-v2-block').profileType,
+      ).toBe('CAMPUS_V2');
+    });
+  });
+
   describe('previewForm', () => {
     it('navigates to the preview page', () => {
       spyOn($location, 'path');
--- a/app/features/eventForm/eventForm.controller.spec.js
+++ b/app/features/eventForm/eventForm.controller.spec.js
@@ -132,23 +132,30 @@ describe('Controller: eventForm', function () {
       scope.$digest();
     });
 
-    it('forces campusV2Question blocks to the CAMPUS_V2 profile type on save', () => {
+    it('forces campusV2Question blocks to CAMPUS_V2 on save, leaving other blocks untouched', () => {
       spyOn(ConfCache, 'update');
       $httpBackend.expectPUT(/^conferences\/.+$/).respond(204, '');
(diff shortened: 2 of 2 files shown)


app/features/eventForm/eventForm.controller.js
```diff
@@ -56,6 +72,7 @@ angular
         }
 
         formSaving = true;
+        normalizeCampusProfileTypes();
```
kegrimes, line 75: [Must-fix] normalizeCampusProfileTypes() mutates the deep-$watch'd $scope.conference inside saveForm; the conference $watch detects the change and fires a second redundant PUT via formSavingTimeout. Could we normalize the outgoing copy or run it once on load instead? [Category: File]

app/features/eventForm/eventForm.controller.js
```diff
@@ -46,6 +46,22 @@ angular
...
+          (page.blocks || []).forEach(function (block) {
+            if (
+              block.type === 'campusV2Question' &&
+              block.profileType !== 'CAMPUS_V2'
+            ) {
+              block.profileType = 'CAMPUS_V2';
+            }
+          });
+        });
+      }
```
kegrimes, line 63: [Mentoring] Mutating a deep-$watch'd model inside its own save handler is re-entrant: the change is detected on the next digest and re-enters saveForm. Prefer normalizing on load or on the outgoing payload for a single-pass save. [Category: File]

kegrimes (commented):
PR adds normalizeCampusProfileTypes() to force campusV2Question blocks to the CAMPUS_V2 profile type on save. Key findings: (1) Must-fix: it mutates the deep-$watch'd $scope.conference inside saveForm, so the first correcting save triggers a second redundant PUT via the debounce timeout. (2) The campusV2Question->CAMPUS_V2 mapping is now hardcoded a third time (already in blockEditor typeToProfile and questionToolbar defaultProfile). (3) Test coverage is a single happy-path case and does not assert single-PUT behavior, so it cannot catch the double-save. No i18n, DOM, payment, or couple/spouse concerns in scope; defensive || [] guards are good.

app/features/eventForm/eventForm.controller.js
```diff
@@ -46,6 +46,22 @@ angular
...
+        (
+          ($scope.conference && $scope.conference.registrationPages) ||
+          []
+        ).forEach(function (page) {
+          (page.blocks || []).forEach(function (block) {
+            if (
+              block.type === 'campusV2Question' &&
+              block.profileType !== 'CAMPUS_V2'
+            ) {
+              block.profileType = 'CAMPUS_V2';
```
kegrimes, line 59: [Reuse] 'campusV2Question'->'CAMPUS_V2' is already mapped in blockEditor.js:228 (typeToProfile) and questionToolbar.js:175 (defaultProfile). Could we extract a shared type->profile map instead of a third literal pair? [Category: Reuse]

app/features/eventForm/eventForm.controller.js
```diff
@@ -46,6 +46,22 @@ angular
...
       var formSavingNotifyTimeout;
 
+      function normalizeCampusProfileTypes() {
+        (
+          ($scope.conference && $scope.conference.registrationPages) ||
+          []
+        ).forEach(function (page) {
+          (page.blocks || []).forEach(function (block) {
+            if (
+              block.type === 'campusV2Question' &&
```
kegrimes, line 56: [Pattern] Hardcoded block-type/profile keys 'campusV2Question'/'CAMPUS_V2' rather than a shared constant. Consistent with the repo today, but a constants module would prevent drift if the key changes. [Category: Pattern]

app/features/eventForm/eventForm.controller.spec.js
```diff
@@ -127,6 +127,31 @@ describe('Controller: eventForm', function () {
...
+          type: 'campusV2Question',
+          profileType: null,
+        });
+      });
+      $httpBackend.flush();
+
+      expect(
+        page.blocks.find((block) => block.id === 'campus-v2-block').profileType,
+      ).toBe('CAMPUS_V2');
+    });
```
kegrimes: [Suggested tests] Add idempotency (already-CAMPUS_V2 untouched), a non-target block left unchanged, and assert exactly one PUT (flush $timeout) to catch the saveForm self-trigger double-save. [Category: File]

app/features/eventForm/eventForm.controller.spec.js
```diff
@@ -127,6 +127,31 @@ describe('Controller: eventForm', function () {
     });
   });
 
+  describe('normalizeCampusProfileTypes', () => {
+    beforeEach(() => {
+      scope.$digest();
```
kegrimes, line 132: [Nice-to-have] This beforeEach scope.$digest() lacks the explanatory comment the saveForm describe uses. Could we note it seeds oldObject so future $watch triggers fire? [Category: File]

app/features/eventForm/eventForm.controller.js
kegrimes: /dismiss it would found out that running on load does not work

app/features/eventForm/eventForm.controller.js
kegrimes: /dismiss skip

app/features/eventForm/eventForm.controller.js
kegrimes: /dismiss skip

app/features/eventForm/eventForm.controller.js
```diff
@@ -46,6 +46,22 @@ angular
       var formSavingTimeout;
       var formSavingNotifyTimeout;
 
+      function normalizeCampusProfileTypes() {
+        (
+          ($scope.conference && $scope.conference.registrationPages) ||
+          []
+        ).forEach(function (page) {
+          (page.blocks || []).forEach(function (block) {
```
canac, line 54: Optional-chaining would simplify this

```suggestion
          $scope.conference?.registrationPages?.forEach(function (page) {
          page.blocks?.forEach(function (block) {
```

app/features/eventForm/eventForm.controller.js
```diff
@@ -46,6 +46,22 @@ angular
...
+            if (
+              block.type === 'campusV2Question' &&
+              block.profileType !== 'CAMPUS_V2'
```
canac, line 57: You could also always set the profile type, which will be a no-op if it's already `CAMPUS_V2`.

```suggestion
```

canac (commented):
This code makes sense! But is it true that campus v2 questions should always be profile questions?

> Unchecking "Cru Profile" when editing the campus question
If we automatically re-check the box when saving, would it be better to hide or disable the Cru Profile checkbox on campus questions to avoid confusion?

This PR also makes it possible to have two blocks with the same campus profile. Is that desirable? `eventForm.controller.js` seems to try to ensure that no blocks have the same profile.

kegrimes:
@canac I am closing this PR in favor of a backend fix. The issue is that campus V2 questions only get their name resolved (`{ id, name }`) when the block's profile type is `CAMPUS_V2`. However, profile types are allowed to be `null` which happens when a user unchecks "Cru profile" checkbox (thanks for helping me realize that). We can't force every campus question to be `CAMPUS_V2` because that would produce a lot of other bugs. The real fix is to switch the enrichment to gate on the block type instead of the profile type which will match how the CSV export is handled.

PR #995, opened 2026-07-08
[No-Jira] Fix CAMPUS_V2 Cru Profile checkbox

kegrimes:
## Description

> The Cru profile checkbox for that question appears to be able to be unchecked, which i don't think should be allowed.

We should have it checked always ([helpscout ticket](https://secure.helpscout.net/conversation/3373373152/1665844/))

## Testing

- Go to event dashboard
- Create an event and add a campus question
- Click "Edit"
- Check that "Cru Profile" checkbox is checked always

## Checklist

### DIFF
--- a/app/scripts/directives/blockEditor/blockEditor.js
+++ b/app/scripts/directives/blockEditor/blockEditor.js
@@ -225,7 +225,7 @@ angular.module('confRegistrationWebApp').directive('blockEditor', function () {
       $scope.typeToProfile.opportunitiesQuestion = 'OPPORTUNITIES';
       $scope.typeToProfile.birthDateQuestion = 'BIRTH_DATE';
       $scope.typeToProfile.campusQuestion = 'CAMPUS';
-      $scope.typeToProfile.campusV2Question = 'CAMPUS_V2';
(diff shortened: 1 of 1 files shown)


app/scripts/directives/blockEditor/blockEditor.js
zweatshirt, line 1: This seems to cause a regression when adding more than one Campus question. The checkbox altogether disappears. However, I noticed in production, that adding a second campus, and then selecting the 'Cru Profile' checkbox produces this error: 'Only one Campus v2 profile block can be used per form' which in itself seems like a bug. I noticed you seemed to mention these issues to Ben in the ticket.

zweatshirt (approved):
Approving since this addresses the immediate issue

app/scripts/directives/blockEditor/blockEditor.js
kegrimes: I think this is the same behavior as adding a second email question.The checkbox would disappear but I do think that error you get in production is intentional. I don't think you can have two questions that are `CAMPUS_V2` - one would have to be `null`. I'm hoping Ben can fix the other issues I was bringing up!

PR #998, opened 2026-08-04
HS-1701446 - Disable double payment in the event that payment is processed but the registration completion fails

zweatshirt:
### HS ticket: 
- https://secure.helpscout.net/conversation/3406039462/1701446

### Issue
- Any error during registration completion, but after payment, leads to the user being able to resubmit, and consequently, the user is re-charged for the registration.

### Implementation
- Blocks successive payments if the user has already been charged for the registration.
- Provides an explicit modal message to the user that the payment has failed and asking them to reach out to the event administrator. Successive attempts to register are allowed. This puts the onus on admin clients, which is not ideal, but is better than charging registering users multiple times.

- *This frontend change is mitigation until a fix is found for the underlying error(s) causing this issue*, and the backend is able to handle payment and registration completion atomically or can use idempotency keys

## Testing

- Without reproducing the bug, I am unsure that this path can be reliably user tested. The jest tests are hopefully sufficient in proving the behavior.

## Checklist

### DIFF
--- a/app/features/reviewRegistration/reviewRegistration.controller.js
+++ b/app/features/reviewRegistration/reviewRegistration.controller.js
@@ -218,15 +218,16 @@ angular
       };
 
       // Display an error that occurred during registration completion
-      function handleRegistrationError(error) {
-        if (!error) {
+      function handleRegistrationError(error, paymentCharged) {
+        if (!paymentCharged && !error) {
           return;
         }
 
         modalMessage.error({
-          message:
-            error.message ||
-            'An error occurred while attempting to complete your registration.',
+          message: paymentCharged
+            ? 'Your card was charged successfully, but there was a problem completing your registration. Please contact the event administrator.'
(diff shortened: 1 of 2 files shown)


canac (approved):
Thanks for implementing this, Zach! I tested by hard-coding this change and verifying in DevTools that it only submits payment request once.

```patch
diff --git a/app/features/reviewRegistration/reviewRegistration.controller.js b/app/features/reviewRegistration/reviewRegistration.controller.js
index 9ed724dc..0a498441 100644
--- a/app/features/reviewRegistration/reviewRegistration.controller.js
+++ b/app/features/reviewRegistration/reviewRegistration.controller.js
@@ -292,6 +292,7 @@ angular
               });
           })
           .then(function () {
+            throw new Error('Registration failed!');
             return registration.completeRegistration(currentRegistration);
           })
           .then(function () {

```

